Issue #127, opened 2026-08-13
Repo hygiene: 19 MB of committed build artifacts, stale sshd directive, off-thesis root docs


scttfrdmn:
## Problem

**19 MB of build artifacts are tracked in git.** `git ls-files` shows compiled Linux binaries committed at the repo root:

```
11196642  integration-test
 7751090  test-broker
```

`.gitignore` covers `/bin/`, `*.so`, and `*.test`, but not bare binary names at the root, so these were added and are now cloned by everyone.

**A stale sshd directive is documented.** `configs/pam/ssh` and `configs/pam/README.md` tell operators to set `ChallengeResponseAuthentication yes`, which OpenSSH renamed to `KbdInteractiveAuthentication` in 8.7 (2021) and removed thereafter. Following the instructions on any current distro yields a config sshd refuses to load — and keyboard-interactive is exactly what the device-flow prompt needs, so this is the difference between the module being usable and not.

**Six large design documents sit in the repo root**, ~180 KB total, several about adjacent projects rather than this one:

- `cloudworkstation_oidc_integration.md` (48 KB)
- `tailscale_oidc_integration.md` (38 KB)
- `oidc_pam_comprehensive.md` (32 KB)
- `oidc_provider_configuration.md` (29 KB)
- `research_computing_oidc.md`
- `oidc_pam_project.md`

They predate the implementation and are not linked from README's documentation index, so a reader cannot tell which documents describe the shipped system (`DEPLOYMENT.md`, `configs/CONFIGURATION-GUIDE.md`) and which are speculative.

## Proposed fix

- `git rm --cached integration-test test-broker` and add them to `.gitignore`. (Purging them from history is a separate, disruptive decision — not proposed here; the objects stay in the pack until then.)
- `ChallengeResponseAuthentication` -> `KbdInteractiveAuthentication` in `configs/pam/ssh` and `configs/pam/README.md`, noting the OpenSSH version split.
- Move the design documents to `docs/design/` with a short index stating their status, and drop the ones that describe other projects.

## Acceptance criteria

- [ ] No build artifacts tracked; a fresh clone is small
- [ ] sshd snippets work on current OpenSSH
- [ ] Repo root contains only current, on-topic documentation

Issue #122, opened 2026-08-13
pam_sm_acct_mgmt rubber-stamps the account stack; shipped configs use `account sufficient`


scttfrdmn:
## Problem

`pam_sm_acct_mgmt` performs no authorization and unconditionally succeeds (`pkg/pam/cgo_bridge.c:420-443`):

```c
// L-5: Authorization is enforced entirely in the auth phase ...
// deployments MUST NOT rely on the PAM `account` stack with pam_oidc as the
// sole gate. Returning PAM_SUCCESS here is intentional and documented.
return PAM_SUCCESS;
```

The comment is right about the intent, but every shipped config then does exactly what the comment forbids. `configs/pam/ssh`:

```
account sufficient  pam_oidc.so config=/etc/oidc-auth/broker.yaml
account required    pam_unix.so
account required    pam_access.so
```

`sufficient` + `PAM_SUCCESS` terminates the account stack successfully, so `pam_unix`, `pam_access`, `pam_nologin`, and `pam_time` below it **never run**. Same pattern in `configs/pam/login`, `configs/pam/su`, `configs/pam/sudo`, and `test-projects/ssh-server/docker/ssh-server/pam.d/sshd`.

## Impact

Account-phase controls that operators reasonably expect to hold are silently disabled wherever these examples are used: expired/locked accounts (`pam_unix`), host/network restrictions (`/etc/security/access.conf`), `/etc/nologin`, and time-of-day restrictions (`/etc/security/time.conf`). `configs/pam/login` even documents time-based access control via `pam_time.so` in the same file where it is bypassed.

Two related documentation defects in the same files:

- `configs/pam/ssh` claims Unix password auth is an emergency fallback, but the `auth requisite pam_deny.so` on the line immediately after `pam_oidc.so` makes `auth required pam_unix.so` unreachable. The stack is fail-closed (good) but the documented recovery path does not exist (bad).
- `password sufficient pam_oidc.so` is meaningless: `pam_sm_chauthtok` always returns `PAM_AUTHTOK_ERR`.

## Proposed fix

- `pam_sm_acct_mgmt` returns **`PAM_IGNORE`**, which is the semantically correct answer for a module with no account-phase opinion: `sufficient` no longer short-circuits, the real account modules run, and if `pam_oidc` is the *only* account module the all-ignored stack fails closed instead of granting access. Rewrite the L-5 comment accordingly.
[...]

Issue #120, opened 2026-08-13
PAM module returns PAM_SUCCESS before the user completes device authorization


scttfrdmn:
## Problem

`Broker.Authenticate` reports `Success: true` **and** `RequiresDevice: true` when it *starts* a device flow — before anyone has authenticated (`pkg/auth/broker.go:394-407`):

```go
return &AuthResponse{
    Success:        true,
    SessionID:      session.ID,
    DeviceCode:     deviceFlow.UserCode,
    DeviceURL:      deviceFlow.DeviceURL,
    RequiresDevice: true,
    ...
```

`pam_sm_authenticate` inspects `success` first and returns immediately (`pkg/pam/cgo_bridge.c:372-378`):

```c
success = json_object_get_boolean(success_obj);

if (success) {
    log_pam_message(LOG_INFO, "Authentication successful for user: %s", username);
    json_object_put(response_obj);
    return PAM_SUCCESS;
}

// requires_device is only examined in the !success branch below — unreachable
```

The `requires_device` branch that follows is dead code, and even if it were reached it only prints instructions and returns `PAM_AUTHINFO_UNAVAIL` with the comment *"For now, require manual retry"*. There is no polling loop anywhere: the broker implements a `check_session` IPC request (`internal/ipc/server.go:346`), and **no client in the repo ever sends one**.

## Impact

With the shipped stack lines (`auth sufficient pam_oidc.so` in `configs/pam/{ssh,login,su,sudo}`), `PAM_SUCCESS` from a `sufficient` module short-circuits the auth stack and the login is granted:

- before the user visits the device URL or touches a passkey,
- before the ID token is verified,
- before **identity binding** (`verifyIdentityBinding`) and **`require_groups`** (`verifyRequiredGroups`) run — both live in the background `pollDeviceAuthorization` goroutine (`pkg/auth/broker.go:738-786`), i.e. after the module has already answered.

Anyone who can reach the PAM stack for a username that passes policy authenticates as that user with no credential. The only thing preventing exploitation today is that the module's compiled-in socket path is wrong (separate issue), so `connect()` fails first.

## Proposed fix

In `pkg/pam/cgo_bridge.c`:

1. Extract the flow into `perform_authentication(pamh, socket_path, username, service, rhost, tty, timeout_s)` so it is callable from a test with `pamh == NULL`; reduce `pam_sm_authenticate` to argument parsing plus a call.
[...]

PR #130, opened 2026-08-13
fix(pam): correct the broker socket path and honor module arguments

### LINKED ISSUE #119: pam_oidc.so hardcodes the wrong broker socket path and ignores its module arguments
## Problem

The PAM module connects to a socket path that the broker never listens on.

`pkg/pam/cgo_bridge.h:25`:
```c
#define SOCKET_PATH "/var/run/oidc-auth-broker.sock"
```

Everything else in the repo uses `/var/run/oidc-auth/broker.sock` — `pkg/config/config.go:407` (`server.socket_path` default), `cmd/pam-helper/main.go:38`, `cmd/oidc-admin/admin.go:127`, `configs/examples/broker.yaml`, all provider examples, and `scripts/install*.sh` (`RUN_DIR=/var/run/oidc-auth`).

`pam_sm_authenticate` calls `connect_to_broker(SOCKET_PATH)` with the macro, so on a stock install `connect()` fails with ENOENT and every login returns `PAM_AUTHINFO_UNAVAIL`.

Separately, `parse_arguments` (`pkg/pam/cgo_bridge.c:300`) only understands `debug` and `config=`. There is no way to override the socket path from the PAM stack, and the arguments the shipped configs actually pass are silently discarded:

- `configs/pam/sudo`: `operation=sudo target_user=%u`
- `configs/pam/su`: `target_user=%u`
- `configs/pam/README.md` documents `operation=`, `target_user=`, and `service=` as supported parameters

None of those are implemented. An operator reading the README will believe they have per-operation authorization that does not exist.

## Impact

The PAM module cannot reach the broker as installed, so OIDC authentication does not work at all in a default deployment. Once the path is corrected, the ordering bug in `pam_sm_authenticate` becomes reachable — these need to land in the right order.

## Proposed fix

- Correct `SOCKET_PATH` to `/var/run/oidc-auth/broker.sock`.
- Extend `parse_arguments` with `socket=<path>` (validated: absolute, shorter than `sizeof(sun_path)`) and `timeout=<seconds>`, and use the parsed value rather than the macro.
- Log unrecognized arguments at `LOG_WARNING` instead of ignoring them silently.
- Correct `configs/pam/README.md` to document only the parameters that exist.

## Acceptance criteria

- [ ] Default build connects to the broker's default socket
- [ ] `socket=` and `timeout=` module args are honored
- [ ] Unknown args produce a syslog warning
- [ ] `configs/pam/README.md` lists only implemented parameters

scttfrdmn:
Fixes #119

> **Stacked on #128** — base is `ci/pam-package-coverage`. Merge #128 first; GitHub will retarget this to `main`.

`pam_oidc.so` has never been able to reach the broker in a default install:

| | path |
|---|---|
| module (`pkg/pam/cgo_bridge.h:25`) | `/var/run/oidc-auth-broker.sock` |
| broker (`pkg/config/config.go:407`) | `/var/run/oidc-auth/broker.sock` |

Every authentication therefore returned `PAM_AUTHINFO_UNAVAIL`. That mismatch is also what has been masking the `PAM_SUCCESS` bypass in #120 — which is why this PR must land *before* that one is fixed, not after.

## Changes

- `SOCKET_PATH` now matches the broker's default, with a comment tying the two together.
- **New `socket=<path>` module argument** for non-default deployments. Validated: must be absolute and must fit in `sockaddr_un.sun_path`, otherwise it is refused with a log line and the default is kept. Silently truncating would connect to a *different* socket.
- `MAX_SOCKET_PATH` is derived from the platform's `sun_path` (108 on Linux, 104 on Darwin/BSD) instead of being hardcoded to 108, with a `_Static_assert` that the compiled-in default fits.
- `connect_to_broker` refuses an empty or overlong path instead of `strncpy`-truncating it.
- `parse_arguments` takes an out-parameter (`pam_oidc_options`) rather than writing globals, and is no longer `static`, so it can be tested.
- **Unrecognized arguments are logged at `LOG_WARNING`.**

## The arguments that never existed

Turning on that warning exposed the fact that the shipped example configs pass parameters the module has never implemented:

- `config=/etc/oidc-auth/broker.yaml` — the module reads no config file; it only talks to the broker. (Also note `broker.yaml` is the *broker's* root-owned config.) Still accepted with a warning so existing PAM stacks keep working.
- `operation=sudo`, `target_user=%u`, `service=ssh` — no such knobs. PAM does not expand `%u` in module arguments either, so `target_user=%u` was passed through literally. The target user and service name already reach the broker via `pam_get_user()` and `PAM_SERVICE`.

Removed from `configs/pam/{ssh,login,su,sudo,common-auth}`, `QUICK-START.md`, `DEPLOYMENT.md`, and `test-projects/ssh-server`. `configs/pam/README.md` now has a table of the two real arguments plus an explicit "arguments that do *not* exist" section, replacing its old list which documented all four of the fake ones.

## Tests

New `pkg/pam/cgo_args_test.go` drives the real C `parse_arguments` through a thin cgo wrapper (cgo is not allowed in `_test.go` files, hence `cgo_args.go`): default path, override, last-wins, relative rejected, empty rejected, overlong rejected, unknown args inert, and a boundary test at exactly `sun_path`-1 vs `sun_path` bytes.

Verified with `make verify-linux`: vet clean, `pkg/pam` tests pass, `golangci-lint run ./...` reports 0 issues.

## Not included

`configs/pam/ssh` still documents `ChallengeResponseAuthentication` (renamed to `KbdInteractiveAuthentication` in OpenSSH 8.7) and still claims `pam_unix.so` is a working fallback when the `auth requisite pam_deny.so` above it makes that unreachable. Those are tracked in #127 and #122.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,30 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed (BREAKING for PAM configs)
+- **(#119) `pam_oidc.so` connected to the wrong socket and ignored its module
+  arguments.** The compiled-in `SOCKET_PATH` was
+  `/var/run/oidc-auth-broker.sock` while the broker listens on
+  `/var/run/oidc-auth/broker.sock`, so a default install could never reach the
+  broker and every authentication returned `PAM_AUTHINFO_UNAVAIL`. The default
+  is now derived from, and matches, the broker's `server.socket_path` default.
+- **(#119)** Added the `socket=<path>` module argument for non-default
+  deployments. It must be absolute and must fit in `sockaddr_un.sun_path`;
+  anything else is rejected with a log line and the default is kept, rather than
+  being silently truncated to a path naming a different socket. `MAX_SOCKET_PATH`
+  is now taken from the platform's `sun_path` (108 on Linux, 104 on Darwin)
+  instead of being hardcoded.
+- **(#119)** Unrecognized module arguments are now logged at `LOG_WARNING`
+  instead of being silently dropped. **The example configs previously passed
+  `config=`, `operation=` and `target_user=`, none of which the module has ever
+  implemented** (and PAM does not expand `%u` in module arguments, so
+  `target_user=%u` was always literal). These have been removed from
+  `configs/pam/*`, `configs/pam/common-auth`, QUICK-START.md, DEPLOYMENT.md and
+  the ssh-server test project; `configs/pam/README.md` now documents the two
+  arguments that do exist (`debug`, `socket=`) and explicitly lists the ones that
+  never did. `config=` is still accepted, with a warning, so existing PAM stacks
+  keep working.
+
 ### Added
 - **(#118) CI coverage for the PAM/cgo packages.** A new `PAM (cgo)` job installs
   `libpam0g-dev`/`libjson-c-dev` and runs `go vet ./pkg/pam ./cmd/pam-module
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -345,16 +345,16 @@ monitoring:
 #### SSH Configuration (`/etc/pam.d/ssh`)
 ```bash
 # Production SSH PAM configuration
-auth    sufficient  pam_oidc.so config=/etc/oidc-auth/broker.yaml service=ssh
+auth    sufficient  pam_oidc.so
 auth    requisite   pam_deny.so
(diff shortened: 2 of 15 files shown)


scttfrdmn:
Full CI passes on this branch: https://github.com/scttfrdmn/oidc-pam/actions/runs/31721955357

Dispatched manually via `workflow_dispatch`, because `ci.yml` only triggers on `pull_request` targeting `main` — a stacked PR gets no automatic run.

scttfrdmn:

Issue #129, opened 2026-08-13
No working end-to-end test harness: integration suite, Dockerfile, and docker-compose are all non-functional


scttfrdmn:
Everything in the repo that claims to test the client and server sides together is broken. Nothing exercises the real path — PAM module → Unix socket → broker → OIDC provider — anywhere, in CI or locally, which is the environmental reason #120/#121/#122 were never caught.

## What's broken

**1. `test/integration` cannot run at all.** After fixing the `token_encryption_key` (done in #128), all three tests fail at broker construction:

```
Failed to create broker: failed to create OIDC provider 'test': failed to create OIDC provider:
  Get "mock://test-provider/.well-known/openid-configuration": unsupported protocol scheme "mock"
```

`test/integration/broker_test.go` was written against a `mock://` issuer no HTTP client can fetch. It is excluded from CI, so this has never been noticed.

**2. `test/Dockerfile.integration`** pins `FROM golang:1.21-alpine` against a `go 1.25.12` module.

**3. `test-projects/ssh-server/docker-compose.yml` does not parse**, and would not work if it did:

```
yaml: construct errors: line 200: mapping key "services" already defined at line 3
```

- `services:` is declared twice; the second block wipes every service definition.
- References `docker/broker/Dockerfile` and `docker/test-runner/Dockerfile` — neither exists (only `docker/ssh-server/Dockerfile` does).
- Mounts `./config`, `./monitoring`, `./logging`, `./keycloak/import` — none exist (the Keycloak realm import actually lives at `test/keycloak/import/test-realm.json`).
- Healthchecks `curl http://localhost:8080/health` on the broker, which has no HTTP server at all — it listens on a Unix socket, and `pkg/metrics/server.go` only serves `/metrics`.
- Pulls in Prometheus, Grafana and Fluentd, none of which are needed to test authentication.

## Proposed fix: two containers, one hermetic device flow

Replace the aspirational stack with a minimal harness under `test/e2e/`, split along the trust boundary as two images:

- **server side** — the broker (`oidc-auth-broker`) with a test `broker.yaml`, plus a **mock OIDC provider** implementing just enough of RFC 8628 (`/.well-known/openid-configuration`, `/device_authorization`, `/token`, `/jwks.json`) with a self-signed RS256 key, so the flow is hermetic and fast: no Keycloak, no browser, and the test can decide exactly when the device code is "approved", and can mint an ID token whose `preferred_username` deliberately does **not** match the local user.
[...]

PR #131, opened 2026-08-13
fix(pam)!: complete the device flow instead of granting PAM_SUCCESS at initiation

### LINKED ISSUE #120: PAM module returns PAM_SUCCESS before the user completes device authorization

scttfrdmn:
Fixes #120. **Stacked on #130** (base `fix/pam-socket-path`) — review/merge #128 → #130 → this. The socket-path fix has to land first or the bypass this PR closes is unreachable in a default install (the module was connecting to a socket the broker never listens on, which is what has been masking it).

## The bypass

`Broker.Authenticate` returns `success: true` **together with** `requires_device: true` when it has merely *started* the device flow (`pkg/auth/broker.go:394`), and `pam_sm_authenticate` tested `success` before it looked at `requires_device`:

```c
success = json_object_get_boolean(success_obj);
if (success) {
    return PAM_SUCCESS;          // <-- before the requires_device check below
}
```

With the shipped `auth sufficient pam_oidc.so`, `PAM_SUCCESS` short-circuits the auth stack. So login was granted the moment the broker replied:

- before the user ever visited the verification URL — no proof of identity at all;
- before the broker bound the OIDC identity to the requested local username (`username_claim`) and enforced `require_groups`, both of which run *afterwards*, in the `pollDeviceAuthorization` goroutine (`pkg/auth/broker.go:738-786`).

The `requires_device` branch below it was therefore unreachable dead code, and it only returned `PAM_AUTHINFO_UNAVAIL` with a `// For now, require manual retry` comment. Nothing in the tree called `check_session`, so the device flow had no completion path in the first place.

## What changed

**`pkg/pam/cgo_bridge.c`**

- The auth phase is extracted into `perform_authentication()`; `pam_sm_authenticate` is now argument parsing + `get_user_info` + a call. It tolerates a `NULL` `pam_handle_t` (no conversation messages), which is what makes it testable from Go.
- `requires_device` is checked **before** `success` is honored, in one `classify_response()` used for both the initial reply and every poll.
- When device authorization is required: display the instructions once, then poll `check_session` until a terminal state. Each attempt opens its own connection, because the broker serves exactly one request per connection and closes. The interval comes from `metadata.polling_interval`, clamped to [1, 60] s (the broker already clamps what it sends to [5, 3600]; this bound only stops a missing or malformed value becoming a busy loop or an unbounded wait).
- `send_check_session_request()` sends `user_id` as well as `session_id` — `Broker.CheckSession` answers `FORBIDDEN` when they disagree, so a poll without it could never succeed.
- **Fail closed.** Refusals, a session the broker no longer has (`SESSION_NOT_FOUND` / `SESSION_EXPIRED` / `FORBIDDEN` — how the broker reports identity mismatch, group denial, poll failure and expiry), and an exhausted budget all deny, with `error_code` mapped to PAM codes mirroring `errorCodeToPAMResult`. Only *failing to reach an opinion* — unreachable broker, unparseable response, missing `session_id` — returns `PAM_AUTHINFO_UNAVAIL`. `json_object_put` on every path.
- New `timeout=<seconds>` argument (default 90, range 10–900). The default sits below sshd's default `LoginGraceTime` of 120 s, so an expired flow is reported as a denial rather than sshd tearing down the connection mid-prompt. Documented in `configs/pam/README.md`, including the "raise both together" case.

**`internal/ipc/validate.go`** — `user_id` is now **required and validated** on `check_session`, `refresh_session` and `revoke_session`. The broker compares it against the session owner to reject cross-user session access, but the validator neither required nor validated it, so an absent `user_id` was compared as the empty string. No production caller sent these before, so nothing breaks.

**`pkg/pam/pam.go`** — the `PAMResultCode` constants are now taken from the PAM headers the module compiles against instead of being copied by hand. `PAMMaxTries` was `24`, which is not a Linux-PAM code at all (`PAM_MAXTRIES` is 11) — returning it would have turned a deliberate rate-limit denial into an unrecognized error. This surfaced because the new tests compare the C module's return value against the Go constant.

## Tests

`pkg/pam/device_flow_test.go` drives `perform_authentication` against a fake broker on a Unix socket that, like the real one, serves one request per connection and then closes — so a module that failed to reconnect between polls could not pass:

- **`TestPerformAuthenticationDeniesWhenDeviceFlowNeverCompletes`** — the bypass regression test. The broker replies `success: true, requires_device: true` and then stays pending; the module must poll and then deny, not grant.
- device flow completed after two polls → `PAM_SUCCESS`, with the poll requests asserted to carry `type`, `session_id` **and** `user_id`.
- each terminal poll error → its mapped PAM code (7 cases).
- already-active session with no device step → `PAM_SUCCESS` from one request, no polling.
- up-front refusals → mapped codes, and never polled (6 cases).
- `PAM_AUTHINFO_UNAVAIL` only for: no broker listening, malformed JSON, no `success` field, connection closed without a reply, `requires_device` without a `session_id`.
- over-long `session_id` refused rather than truncated.

Plus `timeout=` parsing (11 cases, including the assertion that the default stays under `LoginGraceTime`).

## Verification

`make verify-linux` (added in #128) — `go vet ./...`, `go test -race ./pkg/... ./internal/...`, and `golangci-lint run ./...` all green, cgo packages included. `pkg/pam` now takes ~33 s, dominated by the deliberate polling waits in these tests.

Not covered here: the end-to-end sshd path. Scenarios 1–2 of #129 are the container version of the regression test above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,40 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+- **Critical (#120): `pam_oidc.so` granted `PAM_SUCCESS` before device
+  authorization completed.** `Broker.Authenticate` returns `success: true`
+  *together with* `requires_device: true` when it has merely started the device
+  flow, and `pam_sm_authenticate` tested `success` before it looked at
+  `requires_device` — so with the shipped `auth sufficient pam_oidc.so` the auth
+  stack was short-circuited and login granted the moment the broker replied,
+  before the user visited the verification URL and before the broker performed
+  identity binding (`username_claim`) and `require_groups`, which run afterwards
+  in a background goroutine. `requires_device` is now checked first, and
+  `success` alone is no longer treated as a grant.
+- **(#120)** The module now completes the device flow instead of abandoning it.
+  It displays the verification instructions, then polls `check_session` —
+  reconnecting per attempt, since the broker serves one request per connection —
+  at the interval the broker asks for in `metadata.polling_interval` (clamped to
+  [1, 60] s) until the flow completes or the budget expires. Nothing called
+  `check_session` before, so the device flow had no completion path at all.
+- **(#120) Every non-success outcome now fails closed.** A refusal, a session the
+  broker no longer has (`SESSION_NOT_FOUND`/`SESSION_EXPIRED`/`FORBIDDEN` — the
+  broker deletes the session on identity mismatch, group denial and expiry), or
+  an exhausted timeout all deny; broker error codes are mapped to PAM codes
+  mirroring `errorCodeToPAMResult`. Only failing to reach an opinion at all
+  (broker unreachable, unparseable response) returns `PAM_AUTHINFO_UNAVAIL`.
+- **(#120)** `internal/ipc` now *requires* `user_id` on `check_session`,
+  `refresh_session` and `revoke_session`. The broker compares it against the
+  session owner to reject cross-user access, but the validator neither required
+  nor validated it, so an absent `user_id` was compared as the empty string.
+- **(#120)** `PAMMaxTries` was `24`, which is not a Linux-PAM result code
+  (`PAM_MAXTRIES` is 11); returning it would have turned a deliberate
+  rate-limit denial into an unrecognized error. The `PAMResultCode` constants are
+  now taken from the PAM headers the module is compiled against instead of being
+  copied by hand, since the values are not portable between Linux-PAM and
+  OpenPAM.
+
 ### Fixed (BREAKING for PAM configs)
 - **(#119) `pam_oidc.so` connected to the wrong socket and ignored its module
   arguments.** The compiled-in `SOCKET_PATH` was
@@ -32,6 +66,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   keep working.
 
 ### Added
+- **(#120)** `timeout=<seconds>` module argument bounding the wait for device
+  authorization (default 90, range 10–900). The default sits below sshd's
+  default `LoginGraceTime` of 120 s so an expired flow is reported as a denial
+  rather than sshd dropping the connection; documented in
+  `configs/pam/README.md`.
 - **(#118) CI coverage for the PAM/cgo packages.** A new `PAM (cgo)` job installs
   `libpam0g-dev`/`libjson-c-dev` and runs `go vet ./pkg/pam ./cmd/pam-module
   ./cmd/pam-helper` plus `go test -race ./pkg/pam/...`; the `Lint` job now lints
--- a/configs/pam/README.md
+++ b/configs/pam/README.md
@@ -92,15 +92,36 @@ auth    sufficient  pam_oidc.so debug
 
 ## PAM Module Parameters
 
-`pam_oidc.so` accepts exactly two arguments:
+`pam_oidc.so` accepts exactly three arguments:
 
 | Argument | Meaning |
 |---|---|
 | `debug` | Log at `LOG_DEBUG` to syslog (`LOG_AUTHPRIV`). Remove in production. |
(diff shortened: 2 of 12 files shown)


scttfrdmn:
Full CI (all jobs, including the new `PAM (cgo)` job) passes on this branch: https://github.com/scttfrdmn/oidc-pam/actions/runs/31721965477

Dispatched manually because `ci.yml` only triggers on `pull_request` targeting `main`, so stacked PRs get no automatic run. Base `fix/pam-socket-path` is also green: https://github.com/scttfrdmn/oidc-pam/actions/runs/31721955357

scttfrdmn:

PR #132, opened 2026-08-13
fix(security): complete the device flow in the Go client path too

### LINKED ISSUE #121: Go client path treats device-flow initiation as authentication success
## Problem

The Go-side client has the same defect as the C module. `PAMModule.AuthenticateUser` (`pkg/pam/pam.go:120-129`):

```go
if authResp.RequiresDevice {
    if authResp.DeviceURL != "" { ...log... }
    if authResp.DeviceCode != "" { ...log... }
    // Device flow initiated; the PAM stack will poll via check_session.
    return nil
}
```

The comment describes a polling step that does not exist anywhere in the repo. `return nil` means success, and `cmd/pam-helper/main.go` turns that into:

```go
log.Info().Msg("Authentication successful")
os.Exit(0)
```

So `oidc-pam-helper -user <name>` exits 0 the moment a device flow is *started*, before anyone authenticates and before the broker runs identity binding or `require_groups` (both in the background goroutine, `pkg/auth/broker.go:738-786`).

## Impact

Any PAM stack or script that gates on `oidc-pam-helper`'s exit status is bypassable with no credential. Same severity as the C-module issue; separate code path, separate fix.

## Proposed fix

Add **`internal/brokerclient`** — a pure-Go client (`net.Dial` + `encoding/json`, no cgo) so it is testable in CI:

- `Client{SocketPath, Timeout}` with `Authenticate`, `CheckSession`, and `AuthenticateAndWait` implementing the same poll/terminal-state logic as the C module, with an injectable clock.
- `Request`/`Response` types matching `internal/ipc`.

Then rewrite `PAMModule.AuthenticateUser` to delegate to it and return `*PAMAuthFailure` on denial *and* on poll timeout, keeping the C bindings only for syslog. `cmd/pam-helper` then exits non-zero on denial with no change to its own logic.

## Acceptance criteria

- [ ] `AuthenticateUser` returns an error while a device flow is still pending, and `nil` only once the session is active
- [ ] `oidc-pam-helper` exits non-zero on denial and on timeout
- [ ] Table-driven tests against a fake broker socket, running in CI (no cgo in the tested package)

scttfrdmn:
Fixes #121

**Stacked on #131** (base is `fix/pam-device-flow-bypass`, not `main`). Review that one first; this PR's diff is only its own commit.

## The bug

`PAMModule.AuthenticateUser` returned nil as soon as the broker answered `success: true` — which `Broker.Authenticate` does the moment it *starts* the device authorization flow, alongside `requires_device: true`. Nothing waited for the user to visit the verification URL, and nothing waited for identity binding (`username_claim`) or `require_groups`, both of which run afterwards in a background goroutine. `cmd/pam-helper` then exited 0, so a `pam_exec`-style integration granted the login on an identity nobody had proved.

This is #120 on the other client path. #131 fixed the C module; without this, the Go path is still open.

## What changed

**New `internal/brokerclient`** — the broker IPC protocol and the device-flow completion logic in one place, **pure Go, no cgo**. That is the point: the equivalent logic in `cgo_bridge.c` can only be compiled and tested on a machine with PAM headers, so it went five releases without a test. This package is exercised by `go test` on every platform.

- dials per request, because the broker serves exactly one request per connection and then closes — a client that reuses the connection cannot pass the tests
- polls `check_session` carrying `user_id` (the broker answers `FORBIDDEN` when it does not match the session owner)
- honours `metadata.polling_interval`, clamped to [1, 60] s so a missing or malformed value is neither a busy loop nor an unbounded wait
- shortens the final sleep to whatever remains, so the wait never overshoots its budget
- distinguishes a broker **decision** (`DenialError`, `TimeoutError` — both denials) from a **failure to obtain one** (transport/parse errors). Only the latter becomes `PAM_AUTHINFO_UNAVAIL`; an unfinished flow is never a success.

**`pkg/pam/pam.go`** — `AuthenticateUser` delegates to it (via a new `AuthenticateUserContext`), logging the verification URL and code through the existing syslog bindings, and returns `*PAMAuthFailure` on refusal and on an exhausted timeout.

**`cmd/pam-helper/main.go`** — `-timeout` now bounds the authentication itself (default 90 s, matching the module's `timeout=`); the outer watchdog fires 10 s later, so running out of budget is reported as a denial rather than the process being killed mid-flow.

**Login-type classification** — the C module and the Go client disagreed, so the broker could apply different per-login-type policy to the same login depending on which client ran. C matched `gdm`/`lightdm` as *substrings*, never matched `sddm`, and tested the TTY before the service (so a display manager on `tty1` came out `console`). It is now one documented function, pinned to `GetLoginType` by a test over 14 service/TTY pairs.

## Tests

`internal/brokerclient/client_test.go` runs a fake broker on a Unix socket with a **virtual clock**, so a 30 s polling budget costs no wall-clock time and the elapsed time the loop observes is exactly what it asked for:

- **never completes → `*TimeoutError`, after exactly 7 requests** (1 authenticate + 6 polls at 5 s) — the headline regression test
- succeeds after approval; asserts `OnDeviceFlow` fired and every poll carried `type`, `session_id` and `user_id`
- terminal poll errors (`SESSION_NOT_FOUND`, `SESSION_EXPIRED`, `FORBIDDEN`, `DEVICE_FLOW_FAILED`, `POLICY_DENIED`) → denial, stopping at the first one
- an already-active session grants without polling
- five up-front refusals deny without a single poll; a refusal that *also* sets `requires_device` is still a refusal
- five transport failures (no broker, malformed JSON, closed without a reply, device flow with no `session_id`, broker gone by poll time) are **not** reported as denials
- context cancellation, poll-interval clamping (7 cases), and no budget overshoot (elapsed == exactly 30 s)

## Verification

`make verify-linux` (vet + `go test -race` + golangci-lint on `./...`, in the container, since the cgo packages cannot build on macOS):

```
ok  github.com/scttfrdmn/oidc-pam/pkg/pam               33.060s
ok  github.com/scttfrdmn/oidc-pam/internal/brokerclient  1.018s
...
0 issues.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,29 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Security
+- **Critical (#121): the Go client path also treated device-flow initiation as
+  authentication success.** `PAMModule.AuthenticateUser` returned nil as soon as
+  the broker replied `success: true`, which it does the moment it *starts* the
+  device flow, and `oidc-pam-helper` exited 0 on that — so `pam_exec`-style
+  integrations granted the login before the user had proved anything, exactly as
+  the C module did (#120). It now waits for the flow to complete and returns
+  `*PAMAuthFailure` on refusal and on an exhausted timeout.
+- **(#121)** New `internal/brokerclient`: the broker IPC protocol and the
+  device-flow completion logic in one **pure-Go, cgo-free** package, so the
+  wait/poll/deny behaviour is covered by `go test` on every platform rather than
+  living only in C that a developer laptop cannot compile. It dials per request
+  (the broker serves one request per connection), polls `check_session` with
+  `user_id`, honours `metadata.polling_interval` clamped to [1, 60] s, never
+  overshoots its budget, and distinguishes a broker *decision* (`DenialError`,
+  `TimeoutError`) from a failure to reach one (transport/parse errors → only
+  these become `PAM_AUTHINFO_UNAVAIL`).
+- **(#121)** The C module and the Go client disagreed on login-type
+  classification, so the broker could apply different per-login-type policy to
+  the same login depending on which client ran: the C side matched
+  `gdm`/`lightdm` as substrings, never matched `sddm`, and tested the TTY before
+  the service (classifying a display manager on `tty1` as `console`). The
+  classification is now a single documented function pinned to `GetLoginType` by
+  a test.
 - **Critical (#120): `pam_oidc.so` granted `PAM_SUCCESS` before device
   authorization completed.** `Broker.Authenticate` returns `success: true`
   *together with* `requires_device: true` when it has merely started the device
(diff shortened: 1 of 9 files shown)


scttfrdmn:
CI: `ci.yml` only triggers on pull requests targeting `main`, so a stacked PR gets no automatic run. Dispatched manually — **success**: https://github.com/scttfrdmn/oidc-pam/actions/runs/31722769367

scttfrdmn:

PR #133, opened 2026-08-13
fix(security): pam_sm_acct_mgmt has no opinion; stop rubber-stamping the account stack

### LINKED ISSUE #122: pam_sm_acct_mgmt rubber-stamps the account stack; shipped configs use `account sufficient`

scttfrdmn:
Fixes #122

**Stacked on #132** (base is `fix/go-client-device-flow`). Review that one first; this PR's diff is only its own commit.

## The bug

`pam_sm_acct_mgmt` did no account-phase authorization — that is all done in the auth phase, by the broker — but it returned `PAM_SUCCESS`, and the shipped example configs listed the module as:

```
account sufficient  pam_oidc.so
account required    pam_unix.so
account required    pam_access.so
account required    pam_time.so
```

A `sufficient` module that returns success **ends the phase**. So every account check after that line never ran: `pam_time`, `pam_nologin`, `pam_access`, account expiry, `pam_unix`'s shadow checks. Time-of-day restrictions, `/etc/nologin`, network restrictions in `access.conf`, expired accounts — all silently bypassed for every user, on `ssh`, `login`, `su` and `sudo` alike. The module was answering a question it was never asked.

## The fix

`PAM_IGNORE`. That is the honest answer for a module with no account-phase opinion: PAM does not count an ignored module toward the stack's result, so the modules after it decide, and unlike `PAM_SUCCESS` it cannot short-circuit a `sufficient` entry. If `pam_oidc.so` were the *only* account module, an all-ignored stack yields `PAM_PERM_DENIED` — it fails closed rather than admitting everyone.

The verdict is its own function (`acct_mgmt_verdict`) so `TestAcctMgmtHasNoOpinion` can pin it, with an explicit failure message if it ever regresses to `PAM_SUCCESS`. `PAMIgnore` was also added to the `PAMResultCode` constants, derived from the header like the rest.

## Config changes

- `account sufficient pam_oidc.so` → **`account optional`** in `configs/pam/{ssh,login,su,sudo}`, the DEPLOYMENT.md snippets, `test-projects/ssh-server/{README.md,docker/ssh-server/pam.d/sshd}`. With the `PAM_IGNORE` fix a `sufficient` line no longer short-circuits, but as written it still advertises an authorization the module does not perform.
- `password sufficient` → **`password optional`**: `pam_sm_chauthtok` returns `PAM_AUTHTOK_ERR` (change passwords at the IdP), and as `required` it would block password changes for local accounts too.

## The fake fallback

Four of these configs carried

```
auth    sufficient  pam_oidc.so
auth    requisite   pam_deny.so
auth    required    pam_unix.so try_first_pass    # <- never reached
auth    optional    pam_group.so                  # <- never reached
```

`requisite` returns to the application **immediately** on failure, so nothing after `pam_deny.so` runs. The comments in `configs/pam/ssh`, `su` and `sudo` nonetheless described that line as the emergency Unix fallback ("EMERGENCY ACCESS: ... 1. Unix password authentication"), which is the kind of thing an operator only discovers when they need it. The dead lines are gone and the comments corrected: these stacks are OIDC-only, and if OIDC fails the login is refused.

`configs/pam/README.md` gains a per-phase table plus two sections — why `account optional` and never `sufficient`, and why nothing may follow `pam_deny.so` — including the exact edit to enable a genuine Unix fallback and the trade-off it carries (a user with a local password then bypasses every OIDC policy). DEPLOYMENT.md gains a matching "PAM stack semantics" section. Both point out that sshd's public-key path does not consult the PAM auth stack at all, which is why it is the usual break-glass route.

`configs/pam/login` was left as-is on this point: there the `requisite` module is `pam_nologin.so`, which normally succeeds, so its `pam_unix.so` line is genuinely reachable.

## Verification

`make verify-linux` (vet + `go test -race` + golangci-lint on `./...` in the container):

```
ok  github.com/scttfrdmn/oidc-pam/pkg/pam  34.078s
...
0 issues.
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Security
+- **High (#122): `pam_sm_acct_mgmt` rubber-stamped the account stack.** It
+  returned `PAM_SUCCESS` while performing no account-phase authorization at all,
+  and the shipped configs listed it as `account sufficient pam_oidc.so`. A
+  `sufficient` module that succeeds ends the phase, so that combination silently
+  disabled every account check after it — `pam_time`, `pam_nologin`,
+  `pam_access`, account expiry, `pam_unix`'s shadow checks — for every user. It
+  now returns `PAM_IGNORE`: PAM does not count an ignored module toward the
+  stack's result, so the modules after it decide, and a stack where *every*
+  module ignores yields `PAM_PERM_DENIED` rather than admitting everyone.
 - **Critical (#121): the Go client path also treated device-flow initiation as
   authentication success.** `PAMModule.AuthenticateUser` returned nil as soon as
   the broker replied `success: true`, which it does the moment it *starts* the
@@ -65,6 +74,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   OpenPAM.
 
 ### Fixed (BREAKING for PAM configs)
+- **(#122)** The example configs (`configs/pam/{ssh,login,su,sudo}`, the
+  DEPLOYMENT.md snippets, `test-projects/ssh-server`) now use `account optional
+  pam_oidc.so` and `password optional pam_oidc.so`. **Existing `/etc/pam.d/*`
+  files that say `account sufficient pam_oidc.so` should be changed to
+  `optional`**: with the `PAM_IGNORE` fix above such a line no longer
+  short-circuits, but as written it still advertises an authorization the module
+  does not perform. `password` is `optional` because `pam_sm_chauthtok` returns
+  `PAM_AUTHTOK_ERR` (change passwords at the IdP) and as `required` would block
+  password changes for local accounts too.
+- **(#122)** Removed unreachable `auth required pam_unix.so` / `auth optional
+  pam_group.so` lines that followed `auth requisite pam_deny.so` in the `ssh`,
+  `su`, `sudo` and ssh-server-test configs, and corrected the comments — in
+  several places the docs described that dead line as the emergency Unix
+  fallback. `requisite` returns to the application immediately, so nothing after
+  `pam_deny.so` ever runs: those stacks are OIDC-only. `configs/pam/README.md`
+  and DEPLOYMENT.md now explain the phase semantics, how to enable a real
+  fallback and what it costs, and why SSH public keys are the usual break-glass
+  path.
 - **(#119) `pam_oidc.so` connected to the wrong socket and ignored its module
   arguments.** The compiled-in `SOCKET_PATH` was
   `/var/run/oidc-auth-broker.sock` while the broker listens on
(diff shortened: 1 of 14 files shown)


scttfrdmn:
CI does not run automatically on this PR: `.github/workflows/ci.yml` triggers on `pull_request: branches: [main]`, and this PR targets `fix/go-client-device-flow` (it is stacked). Dispatched manually against the branch instead:

https://github.com/scttfrdmn/oidc-pam/actions/runs/31723274178 — **success**.

scttfrdmn:

PR #134, opened 2026-08-13
fix(auth): store tokens in the encrypted TokenManager, not on the session

### LINKED ISSUE #123: Encrypted TokenManager is never called; refresh tokens are stored in plaintext on the session
## Problem

`TokenManager` implements an encrypted in-memory token store — AES-256-GCM via `pkg/security/encryption.go`, a fingerprint index, constant-time ownership comparison (L-15), and a cleanup goroutine. `TokenManager.StoreToken` (`pkg/auth/token_manager.go:93`) has **no callers**. The broker only ever calls `tokenManager.Start` and `.Stop`.

Instead, `pollDeviceAuthorization` puts the OAuth2 refresh token straight onto the session in cleartext (`pkg/auth/broker.go:794`):

```go
updated.RefreshToken = token.RefreshToken
```

against `Session.RefreshToken string // OAuth2 refresh token for session renewal` (`pkg/auth/broker.go:59`), and `RefreshSession` reads it back from there. Access tokens and ID tokens are simply discarded.

So `security.token_encryption_key` — mandatory since v0.4.0, validated at startup, generated by `oidc-admin gen-key` — currently encrypts nothing. `TokenManager.RefreshToken` is also a stub that returns `"token refresh not implemented"`.

## Impact

README.md and SECURITY.md both advertise "AES-256-GCM token encryption" as a delivered, audited feature. In reality long-lived refresh tokens sit in plaintext in the broker's heap for the lifetime of every session, so anything that can read broker memory (core dump, `/proc/<pid>/mem`, a swap-backed page, a future heap-exposure bug) yields directly reusable IdP credentials. The encryption code exists and is tested — it is simply not on the live path.

## Proposed fix

- Replace `Session.RefreshToken` with `Session.TokenID string`.
- Have `StoreToken` return the generated token ID, and call it from `pollDeviceAuthorization` after the identity-binding and group checks pass. Keep only the ID on the session.
- `RefreshSession` retrieves the refresh token via `tokenManager.GetToken(session.TokenID)`.
- `RevokeSession` and the session-expiry path call `RevokeSessionTokens(sessionID)`.
- Delete the unimplemented `TokenManager.RefreshToken` stub (the broker refreshes through `OIDCProvider.RefreshToken`).

## Acceptance criteria

- [ ] No plaintext refresh token is retained on `Session`
- [ ] Tokens are stored encrypted and retrievable only via the token manager
- [ ] Revoking or expiring a session removes its tokens
- [ ] Tests assert both the absence of plaintext and the revocation behavior
- [ ] The README/SECURITY encryption claim becomes accurate

scttfrdmn:
Fixes #123.

**Stacked on #133** (base is `fix/acct-mgmt-fail-closed`, not `main`) so the diff shows only this change. Review/merge order: #128 → #131 → #132 → #133 → this.

## The problem

`NewTokenManager` is constructed and started at broker startup, and `StoreToken` does encrypt with AES-256-GCM — but **nothing ever called it**. Grep for `StoreToken` before this PR and the only hits are the definition and its tests. So:

- the store was permanently empty, and the "AES-256 encryption for token storage" claim in README/SECURITY described encryption of nothing;
- the actual tokens lived in `Session.AccessToken` / `Session.RefreshToken` / `Session.IDToken` as plaintext strings, on a struct that is copied by value, passed around and logged;
- the refresh token is the one that matters here: it is a standing credential that outlives the access token it renews.

Two related gaps: `RevokeSession` dropped the session but left its tokens in the store until they expired on their own (which can be long after the session), and the same for the expiry sweep. And `TokenManager.RefreshToken` was a stub that validated its arguments and returned `"token refresh not implemented"` — no callers, but it advertised a capability that did not exist.

## The change

`Session` now carries `TokenID` instead of the tokens. `TokenID` is a store lookup key with no value on its own; `TokenFingerprint` stays because it is a hash.

- `pollDeviceAuthorization`, after the identity and `require_groups` checks pass, calls `b.tokenManager.StoreToken(token, session.UserID, session.ID)` and keeps only the ID. **If the store fails, the authentication is denied** — `TOKEN_STORE_FAILED`, recorded in metrics and written as an `authentication_denied` audit event, with the session removed. Completing an authentication whose tokens the broker cannot later produce would just move the failure somewhere less diagnosable.
- `RefreshSession` fetches the refresh token via `GetToken(session.TokenID)`, uses it for the duration of the call, stores the token the provider returns, points the session at the new ID and revokes the pre-refresh entry. Providers rotate refresh tokens; the old one should not linger.
- `RevokeSession` and the expiry sweep call `RevokeSessionTokens(sessionID)`.
- `StoreToken` returns `(tokenID, error)`.

The 5-minute `sessionCleanup` ticker body is extracted into `expireSessions(now time.Time)` so both expiry conditions (absolute and idle) are testable without waiting on a ticker. It takes `now` rather than calling `time.Since` internally, and nil-guards `tokenManager`/`auditLogger` because a number of tests build `&Broker{...}` literals directly.

Deleted: the `TokenManager.RefreshToken` stub, and `TokenManager.config`, which was unused once the stub went.

## Tests

`pkg/auth/session_tokens_test.go`:

- `TestSessionHoldsNoTokenMaterial` — reflects over `Session` and fails on any field whose name contains `Token` other than `TokenID`/`TokenFingerprint`. Structural rather than a comment, because a plaintext token field on `Session` is exactly the bug that was there.
- `TestRevokeSessionDestroysItsStoredTokens`, `TestExpiredSessionsHaveTheirTokensRevoked`, `TestIdleSessionsHaveTheirTokensRevoked` — the token is gone from the store afterwards, and a live session's token is left alone.
- `TestRevokeSessionForWrongUserLeavesTokens` — a refused cross-user revocation destroys nothing.

`TestBrokerRefreshSessionSuccess` now pre-stores an old refresh token and asserts the session ends up pointing at a *new* token ID, that the new entry holds the refreshed refresh token, and that the old entry is gone.

## Verification

`make verify-linux` (Docker: `go vet ./...`, `go test -race ./pkg/... ./internal/...`, `golangci-lint run`) — green, `0 issues`. The cgo packages cannot build on the macOS host, hence the container.

## Note on scope

This makes the encryption-at-rest claim true for tokens the broker holds in memory. It does not add persistence: the store is still in-process, so tokens do not survive a broker restart. That is unchanged behaviour and out of scope here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Security
+- **(#123): OAuth2 refresh tokens were held in plaintext on every session; the
+  encrypted `TokenManager` was never called.** `NewTokenManager` was constructed
+  and started, and `StoreToken` encrypts with AES-256-GCM, but no broker code
+  path ever stored a token — so the documented "AES-256 encryption for token
+  storage" applied to an empty store, while the real credentials sat in
+  `Session.RefreshToken` as plaintext strings in a struct that is copied and
+  passed around freely. The broker now stores tokens through the token manager
+  and keeps only `Session.TokenID`; `RefreshSession` decrypts the refresh token
+  for the duration of the call, stores the rotated token and revokes the
+  pre-refresh entry. A structural test fails if any token-bearing field is added
+  back to `Session`.
+- **(#123)** Revoked and expired sessions now destroy their stored tokens
+  (`RevokeSessionTokens`). Previously the session was dropped while its tokens
+  stayed in the store until the tokens themselves expired — which can be well
+  after the session, and a refresh token outlives the access token it renews. A
+  refused cross-user revocation leaves the owner's tokens intact.
 - **High (#122): `pam_sm_acct_mgmt` rubber-stamped the account stack.** It
   returned `PAM_SUCCESS` while performing no account-phase authorization at all,
   and the shipped configs listed it as `account sufficient pam_oidc.so`. A
@@ -138,6 +154,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `-Wall -Wextra` on the cgo `CFLAGS` for `pkg/pam` and `cmd/pam-module`, and
   `(void)` markers on the genuinely unused PAM entry-point parameters so the C
   bridge compiles warning-free.
+
+### Changed
+- **(#123)** `Session.RefreshToken` is replaced by `Session.TokenID`, and
+  `TokenManager.StoreToken` now returns the ID it generated. Both are internal
+  Go APIs (`pkg/auth`), not wire or config surface.
+
+### Removed
+- **(#123)** `TokenManager.RefreshToken`, a stub that validated its arguments and
+  then returned `"token refresh not implemented"`. Refresh is done by
+  `Broker.RefreshSession` via the provider; the stub had no callers and only
+  suggested a capability that did not exist.
 
 ### Fixed
 - **(#118)** `oidc-pam-helper -version` now prints the build date and git commit
--- a/pkg/auth/broker.go
+++ b/pkg/auth/broker.go
@@ -56,13 +56,18 @@ type Session struct {
 	SourceIP         string
 	UserAgent        string
 	TokenFingerprint string
-	RefreshToken     string // OAuth2 refresh token for session renewal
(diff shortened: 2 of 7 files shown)


scttfrdmn:
CI (manually dispatched, since `ci.yml` only auto-triggers on PRs targeting `main` and this one is stacked): **success** — https://github.com/scttfrdmn/oidc-pam/actions/runs/31723950247

scttfrdmn:

PR #135, opened 2026-08-13
feat(admin): implement the broker's status, sessions_list and keys_list IPC handlers

### LINKED ISSUE #124: oidc-admin status/sessions/keys send IPC request types the broker does not implement
## Problem

Four of the five `oidc-admin` subcommands send IPC request types the broker does not implement.

`cmd/oidc-admin/admin.go` sends:

| Command | `type` sent |
|---|---|
| `status` | `"status"` |
| `sessions` | `"sessions_list"` |
| `keys` | `"keys_list"` |

`Server.handleRequest` (`internal/ipc/server.go:333`) handles only `authenticate`, `check_session`, `refresh_session`, and `revoke_session`; anything else falls to `default` and returns `INVALID_REQUEST_TYPE`. The admin CLI then decodes that error `Response` into its own `StatusResponse`/`SessionListResponse`/`KeyListResponse` structs, which have no overlapping fields, so it prints an empty result rather than reporting the failure:

```
$ sudo oidc-admin sessions
No active sessions.        # even with active sessions
$ sudo oidc-admin status
Version:                   # empty
Uptime:
Status:
```

`health` only stats the socket and config file, so it works by accident. `gen-key` is the one genuinely functional subcommand.

Secondary problem: `oidc-admin` connects as the invoking user, but the IPC server rejects any peer whose uid is not 0 (`verifyPeerCredentials`), so these commands additionally require `sudo` — undocumented.

## Impact

The documented operational tooling silently reports nothing. An operator checking for active sessions or provisioned keys gets a confident, wrong answer.

## Proposed fix

- Change `Server.handleRequest` to return `any` so admin request types can use their own payload shapes rather than being squeezed into `Response`.
- Implement `status`, `sessions_list`, and `keys_list`, returning the shapes `cmd/oidc-admin/admin.go:17-49` already decodes.
- Add `Broker.Status()`, `Broker.ListSessions()`, and `Broker.ListKeys()` sourced from the broker's session map (user, provider, login type, created-at, active state, SSH key).
- Allow the new types in `validateRequest`.
- Document the `sudo` requirement in `DEPLOYMENT.md`.

## Acceptance criteria

- [ ] `sudo oidc-admin status|sessions|keys` return real data from a running broker
- [ ] An unimplemented or malformed request surfaces as an error, not as an empty list
- [ ] `internal/ipc` tests cover the three new request types

scttfrdmn:
Fixes #124. Stacked on #134 (`feat/wire-token-manager`) — review that one first; this PR's diff against `main` will shrink once it merges.

## The bug

Every `oidc-admin` command that talks to the broker was silently broken. The CLI sent `status`, `sessions_list` and `keys_list`; the broker implemented none of them, so each request fell through to `INVALID_REQUEST_TYPE`. The client then decoded that error object into its own response struct — which shares no fields with it — and printed the resulting zero value:

```
$ sudo oidc-admin status
Broker Status: RUNNING          # ...answering INVALID_REQUEST_TYPE
Version:
Uptime:
$ sudo oidc-admin sessions
No active sessions
```

The root cause is drift: the request and response types were declared twice, once on each side, and the two copies disagreed. So the fix is structural, not just "add three handlers".

## What changed

**One definition of the wire types.** New `internal/adminapi`, imported by both the broker and the CLI. Every admin response embeds `adminapi.Error`, so a broker refusal can no longer decode as an empty success, and `TestAdminResponsesRoundTripToClientTypes` decodes the server's own output with `DisallowUnknownFields` — it fails if the two sides drift again.

**The broker can answer.** `pkg/auth/admin.go` adds `Broker.Status()`, `ListSessions()` and `ListKeys()`; `internal/ipc/admin.go` maps them to wire types, keeping the layering the repo already uses (`cmd/oidc-admin` never imports `pkg/auth`). `Server.handleRequest` returns `any`, since these answers have their own shapes rather than being squeezed into an authentication response's fields.

- `status`: version (via `SetVersion`, called from `cmd/broker` with the ldflags value — `pkg/auth` cannot see it), start time, uptime, sorted providers, active vs pending session counts. An unstarted broker reports zero uptime rather than a fabricated one.
- `sessions`: user, provider, login type, creation time, expiry, status. **`pending`** rows are device flows the user never completed — that is what a "login is hanging" report actually looks like, and until now there was no way to see them.
- `keys`: algorithm, size in bits, status, expiry, from the new `KeyManager.ListKeyInfo`.

**Nothing in a listing is a credential** — no tokens, no token IDs, no key material. `TestSessionInfoCarriesNoTokenMaterial` is a reflection guard against `SessionInfo` regaining a token field.

**Truthful listings.** `ListKeyInfo` skips key directories it cannot read but *counts* them, surfaced as `unreadable` on the wire and a warning in the CLI, so a short listing is never mistaken for a complete one. Sessions sort newest-first and keys by username, so repeated runs are diffable.

**A client that can tell failure from silence.** `oidc-admin` no longer decides whether the broker is running by opening a throwaway connection — a broker that accepts connections but cannot answer them was reported as healthy. It sends the real request, distinguishes `errBrokerUnreachable` (prints `STOPPED`) from a broker that answers with an error, and bounds every request with a 10-second deadline so a broker that goes quiet mid-response cannot hang the CLI.

Also: `Session.LoginType`, because the broker applied per-login-type policy without recording which login type the session was opened for.

## Access control

Unchanged: the broker socket accepts uid 0 only, via the existing peer-credential check, so all five commands need `sudo`. `DEPLOYMENT.md` gains an "Administrative CLI" section documenting the commands, the `sudo` requirement, `OIDC_SOCKET_PATH`, and the pending/active/expired status values; the weekly-maintenance block now uses `oidc-admin` instead of the deleted `./test-broker` binary.

## Tests

`internal/adminapi` and `cmd/oidc-admin` are added to CI's vet, test and lint lists (neither was covered before). New tests: routing and payload shape for all three types, `Status` on a started broker, empty listings encoding as `[]` rather than `null`, the drift guard, error-shaped responses, `formatUptime`, `validateRequest` acceptance, `ListKeyInfo` algorithm/size/expiry plus the unreadable-directory path, and a fake-broker suite for the client covering decode, broker error, unreachable broker and the request deadline.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,12 +32,14 @@ jobs:
         run: |
           # PAM/cgo packages are vetted and tested in the separate `pam` job,
           # which installs the PAM and json-c development headers.
-          go vet ./pkg/auth ./pkg/config ./pkg/metrics ./pkg/policy ./pkg/security ./pkg/ssh ./internal/ipc ./cmd/broker ./cmd/oidc-admin
+          go vet ./pkg/auth ./pkg/config ./pkg/metrics ./pkg/policy ./pkg/security ./pkg/ssh \
+            ./internal/adminapi ./internal/brokerclient ./internal/ipc ./cmd/broker ./cmd/oidc-admin
 
       - name: Run tests
         run: |
           go test -race -coverprofile=coverage.out -covermode=atomic \
-            ./pkg/auth ./pkg/config ./pkg/metrics ./pkg/policy ./pkg/security ./pkg/ssh ./internal/ipc
+            ./pkg/auth ./pkg/config ./pkg/metrics ./pkg/policy ./pkg/security ./pkg/ssh \
+            ./internal/adminapi ./internal/brokerclient ./internal/ipc ./cmd/oidc-admin
 
       - name: Check test coverage
         run: |
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -132,6 +132,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   keep working.
 
 ### Added
+- **(#124)** `oidc-admin status`, `health`, `sessions` and `keys` work. The broker
+  now implements the `status`, `sessions_list` and `keys_list` IPC requests the
+  CLI has always sent, backed by `Broker.Status()`, `Broker.ListSessions()` and
+  `Broker.ListKeys()`. `status` reports the broker's version, start time, uptime,
+  configured providers and session counts; `sessions` lists user, provider, login
+  type, creation time and status, including **pending** device flows the user has
+  not completed (a login that looks like it is hanging shows up here, which is
+  usually the question being asked); `keys` lists each managed key's algorithm,
+  size in bits, status and expiry. Listings carry no credentials — no tokens, no
+  token IDs, no key material — and are sorted (sessions newest first, keys by
+  username) so repeated runs are diffable. All of them require root, since the
+  broker socket accepts uid 0 only; DEPLOYMENT.md documents the commands and the
+  `sudo` requirement.
+- **(#124)** `Session.LoginType`: the broker applied per-login-type policy without
+  recording which login type a session was opened for.
 - **(#121)** `oidc-pam-helper -timeout` now bounds the device-flow wait rather
   than only arming a watchdog (default 90 s, matching the PAM module's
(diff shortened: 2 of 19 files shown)


scttfrdmn:
CI (manually dispatched — `ci.yml` only auto-triggers on PRs targeting `main`, and this one is stacked): **success** — https://github.com/scttfrdmn/oidc-pam/actions/runs/31725365947

scttfrdmn:

PR #136, opened 2026-08-13
fix(auth): select the OIDC provider deterministically, by priority

### LINKED ISSUE #125: selectProvider iterates a map and ignores provider priority
## Problem

`Broker.selectProvider` (`pkg/auth/broker.go:662`) returns the first enabled provider found while iterating a Go map:

```go
func (b *Broker) selectProvider(req *AuthRequest, policyResult *PolicyResult) *OIDCProvider {
	// For now, select the first available provider
	for _, provider := range b.providers {
		if provider.Config.EnabledForLogin {
			return provider
		}
	}
	return nil
}
```

Go map iteration order is randomized, so with more than one provider enabled for login the chosen IdP varies per authentication attempt within a single process. Both parameters are unused.

`ProviderConfig.Priority` exists and is parsed (`pkg/config/config.go:59`, `mapstructure:"priority"`) but is never read anywhere in the codebase — it is the obvious intended tiebreaker.

## Impact

In any multi-provider deployment users are routed to an arbitrary IdP per login, which makes the device-flow experience nondeterministic and `priority` a silently ignored configuration knob.

## Proposed fix

Sort the enabled providers by `Priority` (descending) with the provider name as a stable tiebreaker, and return the first. Either use `req`/`policyResult` for provider selection or drop the unused parameters.

## Acceptance criteria

- [ ] Selection is deterministic across repeated calls and process restarts
- [ ] `priority` in `broker.yaml` actually determines which provider is used
- [ ] Test with several providers inserted in varying order

scttfrdmn:
Fixes #125. Stacked on #135 (`feat/admin-ipc-handlers`) — review that first; this PR's diff against `main` shrinks once it merges.

## The bug

`selectProvider` ranged over `b.providers`, a Go map, and returned the first login-enabled provider it happened to hit:

```go
for _, provider := range b.providers {
	if provider.Config.EnabledForLogin {
		return provider
	}
}
```

On a host with more than one login-enabled provider, two consecutive logins with the *same* configuration could be sent to different identity providers at random. And `priority` — parsed into `config.OIDCProvider` since the field was added — was never read by anything.

## Which direction is "higher priority"?

⚠️ **This is where I deviated from the plan.** The plan said sort `priority` *descending*. The repo's own configs say otherwise: `configs/production/broker-enterprise.yaml:37-91` runs `priority: 1, 2, 3, 4` down the list, and the `priority: 2` entry is the one commented **"Backup provider for failover"**. `configs/examples/broker.yaml`, `configs/providers/okta.yaml`'s multi-authorization-server example, and `oidc_provider_configuration.md` all follow the same convention: 1 is the primary.

Sorting descending would have made every shipped multi-provider config select its *least* preferred provider — the enterprise config would have routed all logins to `contractors`. So: **ascending, 1 is most preferred**, consistent with `nice`, MX records, and `ip route metric`.

## The ordering

1. `enabled_for_login: false` (the default) — not a candidate.
2. `verification_only: true` — not a candidate either. See below.
3. `priority` ascending, 1 first.
4. A provider that **omits** `priority` sorts *after* every provider that sets one. Unset is `0`, which would otherwise outrank an explicit `priority: 1`, so forgetting the field cannot promote a provider over your declared primary. Negative values are treated the same way — a typo, not a way to mean "first".
5. Equal priorities are broken by name, so the order is identical on every host and across restarts rather than merely repeatable by luck.

## `verification_only` (in scope, same class of bug)

Another parsed-but-unread field. A provider marked "may confirm an identity but must not be the one a login is issued against" was a perfectly valid login target whenever `enabled_for_login` was also set. Setting both is contradictory config, and this resolves it the safe way: not a login candidate. Flagging it explicitly since it is a small behaviour change beyond the issue's headline — a config with both flags set on a provider will stop issuing logins against it, which is what the field name asks for.

## Signature change

`selectProvider` no longer takes `*AuthRequest` and `*PolicyResult`. It read neither, and a signature that claims to consider the request is part of why the map iteration survived unnoticed. The doc comment says what to reinstate them alongside.

## Tests

`TestBrokerProviderSelection` asserted only that the result was `"provider1 or provider2"` — precisely the non-determinism at issue. It now requires the priority-1 provider, on every one of 20 calls. New `pkg/auth/provider_selection_test.go` covers the four-provider failover ordering, stability across 100 calls with five equal-priority providers, unset and negative priority, name ordering when no priorities are set, both filters, and the no-candidates case. Run with `-count=3 -race`.

## Documentation

The selection rules only existed implicitly in the example configs. `configs/CONFIGURATION-GUIDE.md` gains a "Multiple Providers and `priority`" section stating them, plus the limitation worth knowing before relying on a `priority: 2` entry: **this is a preference, not health-based failover.** The broker does not probe providers and fall through when the preferred one is unreachable, so a login against a down primary fails rather than quietly using the backup. `configs/examples/broker.yaml` gets inline comments to match.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -171,6 +171,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   bridge compiles warning-free.
 
 ### Changed
+- **(#125)** `Broker.selectProvider` takes no arguments. It accepted an
+  `*AuthRequest` and a `*PolicyResult` and read neither, which is part of why it
+  went unnoticed that the body was non-deterministic; nothing about the request
+  influences the choice today.
 - **(#124)** `Server.handleRequest` returns `any`, since the admin requests answer
   with their own shapes rather than squeezing a session listing into the fields of
   an authentication response. `internal/adminapi` and `cmd/oidc-admin` are covered
@@ -186,6 +190,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   suggested a capability that did not exist.
 
 ### Fixed
+- **(#125)** Provider selection is deterministic. `selectProvider` ranged over a
+  Go map and returned the first login-enabled provider it happened to hit, so on
+  a host with more than one such provider, consecutive logins could be sent to
+  different identity providers at random — and `priority` was read from the
+  config file and then never used. Candidates are now ordered by `priority`
+  ascending (**1 is the most preferred**, matching every shipped config, where the
+  primary provider is `priority: 1` and the failover provider is `priority: 2`),
+  then by name so that equal priorities still order identically on every host and
+  across restarts. A provider that omits `priority` sorts *after* every provider
+  that sets one, so forgetting the field cannot promote a provider over the
+  declared primary; negative values are treated the same way. Documented in
+  `configs/CONFIGURATION-GUIDE.md`.
+- **(#125)** `verification_only` is honoured. It was parsed and never read, so a
+  provider marked "may confirm an identity, must not be logged in against" was a
+  valid login target if `enabled_for_login` was also set. It is now excluded from
+  login selection.
 - **(#124) Every `oidc-admin` command that talks to the broker was silently
   broken.** The CLI sent `status`, `sessions_list` and `keys_list`; the broker
   implemented none of them, so each request fell through to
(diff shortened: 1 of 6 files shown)


scttfrdmn:
CI (manually dispatched — `ci.yml` only auto-triggers on PRs targeting `main`, and this one is stacked): **success** — https://github.com/scttfrdmn/oidc-pam/actions/runs/31725749325

scttfrdmn:

PR #137, opened 2026-08-13
chore(auth): run the orphaned-key sweep, delete the unreachable risk engine and auth-code flow

### LINKED ISSUE #126: authorized_keys expiry sweep never runs; pkg/policy and the auth-code flow are dead code
## Problem

Several sizeable pieces of the tree have no callers, and one of them is maintenance that should be running.

**The SSH key expiry sweep never runs.** `AuthorizedKeysManager.RemoveExpiredKeys` (`pkg/ssh/authorized_keys.go:276`) strips `authorized_keys` entries older than the hardcoded 24 h window, and it is invoked from nothing but its own test. The broker's `sessionCleanup` loop (`pkg/auth/broker.go:840`) removes expired sessions but leaves the provisioned keys in place, so `authorized_keys` grows monotonically and OIDC-provisioned keys keep granting access after their session is gone. `ListOIDCKeys` and `BackupAuthorizedKeys` are likewise uncalled.

**`pkg/policy` is entirely dead.** `pkg/policy/risk_engine.go` is 548 lines with 482 lines of passing tests and **zero importers** — the live policy engine is `pkg/auth/policy.go`. It is in the CI package lists, so it consumes CI time to test code that cannot execute in production.

**The authorization-code flow is dead.** `pkg/auth/auth_code_flow.go`'s `StartAuthCodeFlow`/`ExchangeCodeForToken` have no callers; only the device flow is wired up.

## Impact

- Real: provisioned SSH keys are never expired from `authorized_keys`, so access outlives its session. This is a security-relevant consequence, not just tidiness.
- Maintenance: ~1500 lines of code and tests that look load-bearing to a reader (and to an auditor) but cannot run, which is how the `requires_device` bug hid in plain sight.

## Proposed fix

- Call `RemoveExpiredKeys` from the broker's `sessionCleanup` loop for users whose sessions have expired, so the 24 h sweep actually executes. Make the window a config value rather than a constant if it is cheap to do so.
- Delete `pkg/policy` and `pkg/auth/auth_code_flow.go` with their tests, and drop `./pkg/policy` from the CI package lists. (Deleting is preferable to leaving it: it is recoverable from git history, and a reader can no longer mistake it for the live engine.)
- Keep `ListOIDCKeys`/`BackupAuthorizedKeys`/`RestoreAuthorizedKeys` — they are legitimate operator/recovery API and gain a caller from the admin-IPC work — but document them as such.

## Acceptance criteria

- [ ] Expired OIDC keys are removed from `authorized_keys` without operator action
- [ ] `pkg/policy` and the auth-code flow are gone; build, vet, lint, and tests are green
- [ ] No remaining exported function in `pkg/ssh` is uncalled and undocumented

scttfrdmn:
Fixes #126. Stacked on #136 (`fix/deterministic-provider-selection`) — review that first.

## The sweep that never ran

`AuthorizedKeysManager.RemoveExpiredKeys` was written to remove broker-issued keys older than 24 hours from a user's `authorized_keys`. Nothing anywhere called it.

That is more than unused code. Sessions live only in the broker's memory, so **a broker restart orphans every key it had issued** — the `@oidc-pam-<timestamp>` lines stay in `authorized_keys` with no session left to revoke them. Each one is a working credential for whoever holds the matching private key, with no expiry and nothing scheduled to remove it.

The session-expiry pass now sweeps each user whose session just expired:

- **Once per user, not once per session.** The sweep rewrites the file under a lock; three expired sessions for one user is one pass.
- **Best-effort.** A failure is logged and the remaining users are still swept — the caller has other teardown work that has to finish.
- **Fails safe, never early.** A key whose broker comment is missing or unparseable is retained. The user's own keys are never touched.

⚠️ **Known bound, stated in the code and the CHANGELOG:** only users with a session expiring *now* are reached. A user whose keys were orphaned by a restart and who has not logged in since is not swept. Reaching them means walking every home directory on the host from the broker, which is a bigger hammer than this issue calls for — worth deciding separately rather than smuggling in here.

## Deletions

**`pkg/policy`** (`RiskEngine`, 548 lines + 482 lines of tests) — no importers. It is a second, parallel risk-scoring implementation; the live one is `PolicyEngine` in `pkg/auth/policy.go`, which is what the broker calls. Two engines with no shared interface and only one wired up is an invitation to fix the wrong one.

**`pkg/auth/auth_code_flow.go`** (`StartAuthCodeFlow`, `ExchangeCodeForToken`, `AuthCodeFlowState`) — no callers. The authorization code flow needs a browser redirect back to a local listener, which does not exist at a PAM prompt on a headless host; the device authorization grant is the reason this project exists. Recoverable from history if a web flow is ever added.

Consequences: `./pkg/policy` drops out of CI's vet and test lists, and `go mod tidy` demotes `go-jose/go-jose/v4` to indirect — it was a direct dependency only because the deleted auth-code-flow test signed its own ID tokens. Without that, CI's go.mod tidiness check would fail.

## Documentation

`RemoveExpiredKeys` now records where it is called from, what it is for, and that **its 24-hour cutoff is fixed and is not the configured session lifetime** — easy to misread as policy-driven.

`ListOIDCKeys`, `BackupAuthorizedKeys` and `RestoreAuthorizedKeys` are kept and marked operator-facing. They are not dead, but nothing in the authentication path calls them, and the next reader auditing for dead code should not have to work that out again. `RestoreAuthorizedKeys` also notes that it reinstates whatever broker-issued keys were present at backup time.

## Tests

New `pkg/auth/key_sweep_test.go`: the orphaned key removed while a key issued an hour ago and the user's own `ssh-ed25519` line both survive; a live session's user untouched; one sweep for a user with three expired sessions; a failing sweep (missing home directory, invalid username, empty user) not preventing another user's; nil `AuthorizedKeysManager`.


scttfrdmn:
CI (manually dispatched — `ci.yml` only auto-triggers on PRs targeting `main`, and this one is stacked): **success** — https://github.com/scttfrdmn/oidc-pam/actions/runs/31726104214

scttfrdmn:

PR #138, opened 2026-08-13
chore: drop 19 MB of committed binaries, fix the sshd directive, file the design docs

### LINKED ISSUE #127: Repo hygiene: 19 MB of committed build artifacts, stale sshd directive, off-thesis root docs

scttfrdmn:
Fixes #127. Stacked on #137 (`chore/remove-dead-code`) — the last PR in the series; review the stack in order.

## Committed build artifacts

`integration-test` (11.2 MB) and `test-broker` (7.75 MB) were tracked macOS arm64 binaries in the repository root. Every clone paid for them; no build step consumed them.

⚠️ **This does not shrink the history.** The blobs are still reachable from old commits — removing them needs a `git filter-repo`/BFG rewrite and a force-push, which is out of scope here and worth deciding deliberately.

`QUICK-START.md` told the reader to run `./test-broker` twice, once with a block of expected output. A Mach-O arm64 binary cannot execute on the Linux hosts that guide is about, so those steps could never have worked as printed. They now use `sudo oidc-admin status` / `sudo oidc-admin sessions`, which work as of #124, with the real output format.

## `.gitignore` ignored `*.h`

The cgo bridge headers are **source files** (`pkg/pam/cgo_bridge.h`), so a newly added header would be silently skipped by `git add` — in the one component of this project written in C, and the one where the device-flow bypass lived. Removed, with a comment saying why, replaced by the object-file patterns it was presumably reaching for.

The last line was also the single mangled entry `*.c.o.claude/`, which matches nothing. It is now `*.c.o` and `.claude/`.

Separately: `/configs/production/` is ignored while three secret-free templates in it are tracked, so a new file there needs `git add -f`. That is the right default for a directory where real credentials end up — I left the behaviour alone and wrote the reason down, rather than un-ignoring it and risking someone's `broker-prod.yaml` becoming committable.

## `ChallengeResponseAuthentication` → `KbdInteractiveAuthentication`

Deprecated in OpenSSH 8.7; only a legacy alias now. The device-flow prompt reaches the user over keyboard-interactive authentication, so this is precisely the directive that has to be right. Fixed in `QUICK-START.md`, `configs/pam/README.md`, `configs/pam/ssh` and the test SSH server's `sshd_config`, each noting the old name so a reader on an older sshd is not confused.

## `test/Dockerfile.integration`

Pinned `golang:1.21-alpine` — four minor versions behind the `go 1.25.12` directive in `go.mod` — so `docker compose -f docker-compose.test.yml build` either failed outright or quietly spent time downloading a second toolchain. Now `golang:1.25.12-alpine`, matching go.mod and CI.

## Design documents

The six pre-implementation design/positioning documents in the repository root (5,546 lines) move to `docs/design/`. **Nothing deleted** — the plan said to confirm before deleting, so I moved all six, including the two about separate projects (Tailscale, CloudWorkstation), which is the reversible choice.

In the root they outnumbered the actual documentation and read as if they described shipped behaviour; several describe features that were never built, and `oidc_provider_configuration.md` contains YAML that no longer matches the schema. The new `docs/design/README.md` states that plainly, points at the seven documents that *do* describe the current system, and characterises each file. `README.md` gains one line pointing at the directory with the caveat attached. Nothing linked to them.


scttfrdmn:
CI (manually dispatched — `ci.yml` only auto-triggers on PRs targeting `main`, and this one is stacked): **success** — https://github.com/scttfrdmn/oidc-pam/actions/runs/31726452738

scttfrdmn:

Issue #189, opened 2026-08-14
TestServerRateLimitOverSocket has never run in CI and fails when it does: validation precedes rate limiting


scttfrdmn:
Found while reproducing an unrelated CI failure inside a root container, where this test actually executes.

## The test never runs

`internal/ipc/server_test.go:661`:

```go
if runtime.GOOS != "linux" || os.Getuid() != 0 {
    t.Skip("verifyPeerCredentials requires Linux + root; skipping rate limit socket test")
}
```

GitHub Actions runs as the unprivileged `runner` user, so this skips on every CI run this test has ever been part of. It is green because it is absent, not because it passes.

## It fails when it runs

```
--- FAIL: TestServerRateLimitOverSocket (0.10s)
    server_test.go:702: expected RATE_LIMIT_EXCEEDED on 3rd request, got error_code="INVALID_REQUEST"
```

The request the test sends is `{"type":"check_session","user_id":"test-user"}` with no `session_id`. In `Server.handleConnection`, `validateRequest` runs at `internal/ipc/server.go:302` and the rate limiter at `:346` — so the request is rejected as `INVALID_REQUEST` before it is ever counted, and no number of them reaches the limiter. The first two assertions ("must not be rate-limited") pass for the wrong reason, and the third cannot pass at all.

## Two things to decide, not one

1. **The test.** It must send a request that passes validation (a `check_session` with a `session_id`) so it exercises the limiter rather than the validator. Its own construction hid the ordering from the author.
2. **The ordering — the security question.** Malformed requests are currently free: a local peer can spin the accept loop, JSON decode and validation as fast as it can write, and the per-peer limit that exists to bound that work never sees them. The counter-argument is real too — the limiter is keyed on `UserID`/class, and an invalid request may have neither, so there is nothing sound to key on. A defensible answer is to keep validation first but count invalid requests against a per-peer-UID bucket (the peer credential is known from `verifyPeerCredentials` before any parsing), so an unparseable flood is bounded without the limiter having to trust a field from the request body.

Also worth deciding whether root-only tests should run at all: a test that silently skips in every environment the project checks is indistinguishable from a deleted one. Either give CI a privileged job for them (a container step running as root would do) or make `verifyPeerCredentials` injectable so the case runs unprivileged. Until then, the same skip guard hides whatever else it covers.

scttfrdmn:
[...]

PR #190, opened 2026-08-14
fix(audit): write the events still queued when the logger stops

### LINKED ISSUE #188: Flaky: TestAuditLoggerComplianceRequirements loses events when a sibling subtest's HTTP output is retrying
Blocked the merge of #186 (PR for #170) with a failure unrelated to that change; a rerun and a local container run of the same command both pass.

## What happened

`go test -race -coverprofile=coverage.out -covermode=atomic ./...` on ubuntu-latest:

```
--- FAIL: TestAuditLoggerComplianceRequirements (0.32s)
    audit_functional_test.go:307: Compliance audit missing required field: payment_processing
    audit_functional_test.go:307: Compliance audit missing required field: credit_card_data
    audit_functional_test.go:307: Compliance audit missing required field: certificate
{"level":"warn","error":"audit HTTP endpoint returned status 500","attempt":1,...}
{"level":"warn","error":"audit HTTP endpoint returned status 500","attempt":2,...}
{"level":"error","total_dropped":1,"message":"Audit event channel full — events are being dropped"}
```

Run: https://github.com/scttfrdmn/oidc-pam/actions/runs/31768309200/job/94668703489

## Diagnosis

The three "missing required field" complaints are the symptom, `total_dropped:1` is the cause. A sibling case in the same package deliberately drives an HTTP audit output against a server answering 500, and the retry/backoff for those attempts occupies the audit worker. The event channel fills, the overflow strategy discards, and the discarded event is the compliance case's — which then cannot find its fields in the output and blames a missing field.

So the assertion is fine; what is unsound is that a case asserting "every required field is written" shares a logger, a channel and a worker with a case that deliberately blocks that worker.

## Fix

Make the compliance case independent of what any other case is doing to the worker: its own `AuditLogger` with `overflow_strategy: block` (or `sync`), so a slow output delays the test rather than silently dropping the record it is about to assert on. A test that can lose the event it asserts on cannot distinguish "the field is missing" from "the event never arrived" — which is exactly the confusion the log above shows, and the reason this reads as a field bug rather than a dropped event.

Worth checking the whole package for the same shape while in there: any case that writes events and then reads them back needs a non-dropping strategy, since `drop` is the default path for an unrecognised strategy value.

scttfrdmn:
Fixes #188.

Filed as a flaky test; it is a flaky test *and* a data-loss defect in the audit logger, and fixing the second is what makes the first deterministic.

## The defect

`processEvents` selected on `stopChan`/`ctx.Done()` and returned. Whatever was buffered in `eventChan` at that moment was never written — collected unwritten. The events most likely to be in that buffer are the ones logged immediately before shutdown, i.e. the tail of the audit trail. The default overflow strategy is `block` specifically so that load cannot cost a record (L-9); shutdown was discarding the same records for nothing.

`Stop` also closed a channel with no guard, so stopping on a shutdown path *and* in a deferred cleanup — the ordinary shape, and what these tests do — panicked with `close of closed channel`.

## The flake it explains

CI failed twice on #186 with:

```
--- FAIL: TestAuditLoggerComplianceRequirements
    audit_functional_test.go:307: Compliance audit missing required field: financial_data_access
    ... 6 fields
```

Four cases in `pkg/security` logged events, slept a fixed interval, then read the file and asserted on its contents. That is an assertion about the scheduler: when the worker has not been scheduled yet, the case reports the *field it cannot find* rather than the *event that never arrived*, which is why this reads as a field bug. Reintroducing the drain defect reproduces that exact missing-field list on demand — the log line `total_dropped:1` in the first CI failure was a second, independent symptom of the same "events not yet written" state.

I could not reproduce the CI failure locally (5/5 passes for the package, and 2/2 for `go test -race ./...` under `--cpus=2`, on both this branch and main), which is the other reason not to fix it by lengthening the sleep: a longer sleep is still a bet on a machine I cannot measure.

## Changes

- `pkg/security/audit.go`: `drain()` writes what is buffered; called on both worker exits and again in `Stop`, so a logger that was never started or whose worker already returned on a cancelled context still writes what it accepted. `Stop` returning now means every accepted event is on disk. `Stop` is idempotent via `sync.Once`.
- `pkg/security/audit_functional_test.go`: the four read-back cases stop the logger and then read — no sleeps, and no two-second context bounding the worker while the case is still logging.
- `pkg/security/audit_shutdown_test.go`: new.

## Verification

- `TestStopWritesEventsStillQueued` — three events queued with no worker running; without the fix the file is `0 bytes` for all three.
- `TestStopDrainsAfterTheContextIsCancelled` — an event logged between `cancel()` and `Stop`; without the fix, `0 bytes`.
- `TestStopIsIdempotent` — without `sync.Once`: `panic: close of closed channel`.
- Each proven by reintroducing the specific half of the defect it covers, output pasted above.
- `gofmt -l pkg internal cmd` clean, `go build ./...`, `go test -race ./pkg/... ./internal/...` clean, `go test -race -count=3 ./pkg/security/` clean.

Not addressed here: #189 (`TestServerRateLimitOverSocket` never runs in CI and fails when it does).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     and `Stop` drains again afterwards, so a logger that was never started, or
     whose worker had already returned on a cancelled context, still writes what
     it accepted. `Stop` returning now means every accepted event is on disk.
+  - Draining is bounded by a five-second budget, because an output can be
+    arbitrarily slow: one unreachable HTTP sink costs up to three ten-second
+    timeouts plus backoff for a *single* event, and the default buffer holds a
+    thousand, so an unbounded drain could hold shutdown for hours. A shutdown
+    that never finishes is worse than an incomplete trail, so the budget wins —
+    and what it abandons is counted in `DroppedEvents` and logged with the
+    number, since a gap nobody is told about is indistinguishable from nothing
+    having happened.
   - `Stop` is idempotent (`sync.Once`). It closes a channel, and stopping on a
     shutdown path plus a deferred cleanup — the ordinary shape — used to panic
     with `close of closed channel`.
@@ -37,8 +45,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   Coverage: `TestStopWritesEventsStillQueued` (three events queued with no
   worker; the file is empty without the fix), `TestStopDrainsAfterTheContextIsCancelled`
-  (an event logged in the window between cancellation and `Stop`) and
-  `TestStopIsIdempotent` (which panics without the `sync.Once`). Each was run
+  (an event logged in the window between cancellation and `Stop`),
+  `TestStopGivesUpOnASlowOutput` (forty events behind a slow output with the
+  budget shortened: unbounded, `Stop` takes 1.04s against a 50ms budget and
+  counts nothing) and `TestStopIsIdempotent` (which panics without the `sync.Once`). Each was run
   against the defect reintroduced.
 - **Medium (#172): `configs/pam/common-auth` put a 90-second interactive device
   flow in the auth stack of every service on the host.** `common-auth` is the file
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,38 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Security
+- **Medium (#188): stopping the audit logger discarded every record still
+  queued, and a second `Stop` panicked.** `processEvents` selected on its stop
+  channel and returned without looking at the buffer, so the events sitting in
+  `eventChan` when the broker shut down were collected unwritten. The records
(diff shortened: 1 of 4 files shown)


scttfrdmn:
Self-review found a regression in the first commit and `e99162a` fixes it.

The drain I added writes to **all** outputs, including the HTTP one. Against an unreachable sink that is up to `httpAuditMaxRetries` × `httpAuditTimeout` plus backoff — roughly 33s — for a *single* event, and the default buffer holds 1000. So "drain everything at shutdown" could hold a broker shutdown for hours: I traded a lost record for a hang, which is the worse failure.

`e99162a` bounds the drain with a five-second budget, checked *after* each write so the budget can never stop it before making progress, and accounts for what it abandons in `DroppedEvents` plus an error log naming the count and the budget. An audit gap nobody is told about is indistinguishable from nothing having happened.

`TestStopGivesUpOnASlowOutput` covers it — forty events behind a slow output with the budget shortened to 50ms. Without the bound: `Stop took 1.035805458s: the drain is not bounded by the budget` and `abandoned events were not counted`.

Issue #168, opened 2026-08-14
C module hygiene: NULL conv deref, sticky debug global that logs full broker responses, dead prompt_user, and a JSON string accepted as success=true


scttfrdmn:
## What's wrong

Four independent issues in `cmd/pam-module/cgo_bridge_linux.c`, grouped because they
are one PR's worth of work.

**1. NULL dereference crashes sshd's auth child.** `display_message` (`:297`)
dereferences `conv->conv` with no NULL check after `pam_get_item(PAM_CONV, …)` returns
`PAM_SUCCESS`. A service with no conversation function set crashes the auth child.
`show_message` guards `pamh` but not `conv`.

**2. Sticky `debug` global leaks the whole broker response to syslog.**
`debug_enabled` (`:12`, set at `:380-382`) is a process global with no reset, so once
*any* PAM stack in the process passes `debug`, every later authentication in that
process logs at DEBUG. Combined with the shipped `auth sufficient pam_oidc.so debug`
(`configs/pam/ssh:15`, `configs/pam/login:11`), `:275` writes the **entire broker
response** — device code, instructions and all — to `LOG_AUTHPRIV` on every login.
Note `configs/pam/ssh:41` already advises removing `debug`, which suggests it should
not be the shipped default in the first place.

**3. Dead code holding an unguarded `strncpy`.** `prompt_user` (`:314-348`) has no
caller and copies into a caller-sized buffer. Delete it rather than leave it for
someone to wire up.

**4. A JSON string is accepted as boolean true.** `json_object_get_boolean(success_obj)`
(`:511`) returns true for any non-empty JSON *string*, so `"success":"false"` classifies
as success. Harmless against the real broker (Go emits a JSON bool) but it is an
unnecessary widening of the grant condition in the one function whose whole job is to
be strict about it — the same function that carries the #120 comment.

## Fix

1. NULL-check `conv` and `conv->conv` before use; return without displaying.
2. Set `debug_enabled` from the parsed options per invocation, not once per process,
   and drop `debug` from the shipped PAM stacks.
3. Delete `prompt_user`.
4. Require `json_object_is_type(success_obj, json_type_boolean)` in
   `classify_response`.

## Acceptance criteria

- A test invoking the auth path with `PAM_CONV` unset that does not crash.
- A test asserting `{"success":"false", …}` is not classified as success.
- `debug` absent from `configs/pam/*`, and a test that a second authentication in the
  same process does not inherit the first's debug flag.

Issue #160, opened 2026-08-14
The per-UID IPC rate limit is keyed on a UID that is always 0, so one unauthenticated remote client can lock every login out of the host


scttfrdmn:
## What's wrong

The IPC "per-UID" rate limiter is keyed on a peer UID that can only ever be `0`, so
there is exactly one bucket for the whole machine and `max_requests_per_minute` is a
**global** budget shared by every login on the host.

- `verifyPeerCredentials` runs **unconditionally** at
  `internal/ipc/server.go:243` and rejects any peer with `uid != 0`
  (`peer_creds_linux.go:30`).
- The `requirePeerAuth` block that follows (`server.go:252-271`) can therefore only
  ever assign `peerUID = 0`; if it is disabled, `peerUID` stays at its zero value.
- `server.go:279` — `if !s.rateLimiter.AllowRequest(peerUID)`, bucket keyed by uid
  (`internal/ipc/ratelimit.go:57-91`).

Shipped budgets: `30` in `configs/production/broker-enterprise.yaml:226`, `60` in
`configs/examples/broker.yaml:129`, `configs/production/broker-minimal.yaml:69` and
`broker-cloud.yaml:90`.

## Impact — unauthenticated remote attacker makes the host unloginnable

One PAM login costs 1 `authenticate` plus one `check_session` per poll:
`DEFAULT_AUTH_TIMEOUT 90` / `DEFAULT_POLL_INTERVAL 5` ≈ **19 requests**. So the
shipped budget is under 3 concurrent legitimate logins.

An attacker with no credentials opens SSH connections to any syntactically valid
username at ~1/s. After 30 (enterprise) or 60 requests the bucket is empty and every
subsequent request — **including the `check_session` polls of legitimate logins
already in flight** — is answered `RATE_LIMIT_EXCEEDED`, before the request body is
even read (`server.go:279-285`, ahead of the `json.Decode` at `:293`).

`map_error_code` turns that into `PAM_MAXTRIES` (`cgo_bridge_linux.c:487-490`), and
the shipped stack is `auth sufficient pam_oidc.so` followed by `auth requisite
pam_deny.so` (`configs/pam/ssh:15-16`) — no password fallback. Console login uses the
same module (`configs/pam/login:11`). The host becomes unloginnable for the duration
of the attack and refills at only 30–60 tokens/min.

## Why existing defences don't stop it

- `maxConcurrentConnections = 128` and the 30 s deadline bound concurrency, not
  request count.
[...]

Issue #179, opened 2026-08-14
Conform to the oauth2-pam wire protocol spec (version 1): seven concrete gaps at 2ccd9f7


scttfrdmn:
## oauth2-pam now owns the broker↔PAM wire protocol, and there is a spec to conform to

Following up on #149 and #146 with something concrete. `oauth2-pam` has written the broker↔module IPC contract down as a normative, versioned specification and taken ownership of it:

**https://github.com/scttfrdmn/oauth2-pam/blob/main/docs/wire-protocol.md**

The ownership split is deliberate, and it is the part worth agreeing or disagreeing with first: **one project owns the protocol, the other consumes it.** Version numbers are allocated in oauth2-pam, the reference implementation lives in the same repository as the document so drift is a reviewable diff, and a protocol change starts as an issue there. The reason is the table in #149 — we did not just duplicate code, we duplicated a *bypass*, because the contract was unwritten. Two peers with two written copies of it would only have slowed down the next occurrence.

To be clear about scope: nothing about this makes oauth2-pam's *identity* handling normative for you. You are not expected to authenticate like it does. You are expected to speak the same socket.

### What I found reading this repo at 2ccd9f7

I read the code rather than just pointing at the document, so this is a specific list rather than "please conform". Each item cites what I looked at.

**1. There is no `status` field on the wire at all.** `internal/ipc/server.go:54` `Response` has no `Status`; neither does `AuthResponse` in `pkg/pam/protocol.go`. The concept exists but is buried in the wrong place — `pkg/auth/broker.go:481-484` puts `"status": "pending"` inside `Metadata map[string]interface{}`, and only on the check path. Promoting that to a top-level field is a smaller change than it looks.

Note `SessionStatusPending`/`Active`/`Expired` in `pkg/auth/admin.go` are a *different* vocabulary for the admin view — version 1's authorized/pending/denied/expired/error is the auth wire, and `active` is not one of them.

**2. `success: true` means *pending* on this wire. This is the one that matters.** `pkg/auth/broker.go:441-447` and `478-480` set `Success: true` together with `RequiresDevice: true` for a started or not-yet-active flow, and `cmd/pam-module/cgo_bridge_linux.c:521-532` tests `requires_device` *before* honouring `success`. The comment there is exactly right about why, and the guard works.

But version 1 says `success == true` **iff** `status == "authorized"`. So the two projects currently disagree about the single most safety-critical field on the wire, and the cross-talk is worth spelling out:

[...]

PR #192, opened 2026-08-14
fix(pam): stop shipping a common-auth stack that every service includes

### LINKED ISSUE #172: configs/pam/common-auth puts a 90-second device flow in every service's auth stack, including sudo, su, cron and the display manager
## What's wrong

`configs/pam/common-auth` puts `auth sufficient pam_oidc.so` in the file every
Debian/Ubuntu service `@include`s. Every one of `su`, `sudo`, `gdm`/`sddm`, `polkit`,
`login`, cron's PAM session and anything else in the stack now begins with a module
that wants to print a QR code and wait up to 90 seconds
(`cmd/pam-module/cgo_bridge.h`, `DEFAULT_AUTH_TIMEOUT 90`) for a human with a phone.

Several of those callers can never satisfy a device flow: no controlling terminal, a
conversation function that discards `PAM_TEXT_INFO`, or a graphical prompt that will
render a block of ASCII art. `display_message` uses `PAM_TEXT_INFO`, so whether the
user ever sees the verification URL is a property of each service's conversation
function — unverified outside sshd, which is the only service the e2e harness covers.

The sister project `oauth2-pam` reached the related conclusion that `PAM_TEXT_INFO` is
itself the wrong primitive — OpenSSH buffers informational messages and may not flush
them until authentication returns — and uses `PAM_PROMPT_ECHO_ON` ("Press Enter once
you have approved").

Related: `README.md:233-241` claims **Stable** for Console and GUI across five
distributions. The harness tests exactly one service.

## Impact

A `sudo` that hangs for 90 seconds and then falls through, a display manager showing
ASCII art, and cron jobs paying the timeout. Availability and usability, not a bypass —
`pam_deny.so` still fails closed.

## Fix

1. **Ship no `common-auth`.** Ship `sshd` only, with the break-glass section it already
   has (which is correct and well-written and should stay).
2. If console login is genuinely supported, ship a separate `login` file, and gate the
   module on an interactive-capable service: it already receives `PAM_SERVICE` and
   `PAM_TTY`, so it can return `PAM_IGNORE` immediately for callers that cannot do a
   device flow rather than burning 90 seconds.
3. Consider `PAM_PROMPT_ECHO_ON` instead of `PAM_TEXT_INFO`, matching `oauth2-pam`.
4. Demote the Console/GUI **Stable** claims in `README.md:233-241` to what is actually
   tested, and drop the "Subsequent access uses cached SSH key" line
   (`README.md:165`) until that is true.

## Acceptance criteria

- `configs/pam/common-auth` is gone, or carries a prominent warning and is not
  referenced by any install path.
- A test or e2e case asserting the module returns `PAM_IGNORE` promptly for a
  non-interactive service instead of waiting out the timeout.
- The README platform table matches what CI exercises.

scttfrdmn:
Fixes #172.

## What was wrong

`configs/pam/common-auth` shipped `auth [success=2 default=ignore] pam_oidc.so`, and
`common-auth` is the file every Debian/Ubuntu PAM service `@include`s. An operator who
followed the install instructions in `configs/pam/README.md` ("For system-wide (be
careful!)") put an interactive device flow at the front of the auth stack of `su`,
`sudo`, `gdm`/`sddm`, `polkit`, `login` and `cron` as well as `sshd`. Each of those then
waited up to `timeout` seconds — 90 by default (`DEFAULT_AUTH_TIMEOUT`, and
`parse_arguments` accepts `timeout=` in `[10, 900]`) — for a human with a phone, and
several of them can never satisfy a device flow at all: no controlling terminal, a
conversation function that discards `PAM_TEXT_INFO`, or a graphical prompt that renders
the QR code as a block of ASCII art.

Nothing was admitted that should not have been — `pam_deny.so` still fails closed. What
broke was the machine: `sudo` hanging for 90 s, the display manager showing ASCII art,
cron jobs paying the timeout. And because the shipped stacks have no password fallback
(#160), an auth stack that cannot complete is an operator locked out of their own host.
That is the failure mode this PR is shaped around.

`scripts/install.sh` was also an install path referencing it: it inserted
`@include common-auth` at line 2 of `/etc/pam.d/sshd`, which duplicated the
distribution's own include further down the same file and, on a host carrying the old
`common-auth` advice, ran the device flow twice.

## What changed

- **`configs/pam/common-auth` is deleted.** The per-service files — `ssh`, `login`,
  `su`, `sudo` — are what this project ships. Each header now says it belongs in that
  one service's `/etc/pam.d` file and never in the host-wide `common-auth`/`system-auth`
  stack, and the three CI does not exercise say so.
- **`scripts/install.sh`** no longer inserts `@include common-auth`, and no longer runs
  `sed` against an `/etc/pam.d/sshd` that does not exist (which aborted the install
  under `set -e`). It still only adds `auth sufficient pam_oidc.so`, with no
  `pam_deny.so` after it, so the rest of the existing sshd stack decides when OIDC is
  unavailable — an install must not be able to lock the operator out.
- **`configs/pam/README.md`** loses the system-wide `cp` and gains a "Why no host-wide
  stack is shipped" section plus a table of which stack CI exercises (`ssh`, via
  `test/e2e`) and which are untested examples. It also tells anyone who installed the
  old file to take `pam_oidc.so` back out of `/etc/pam.d/common-auth`.
- **`DEPLOYMENT.md`** and **`QUICK-START.md`** say the same where they tell an operator
  to edit a PAM file. The `QUICK-START.md` example additionally still carried two
  defects fixed in `configs/pam` long ago — an unreachable `auth required pam_unix.so`
  after `requisite pam_deny.so`, and `account required pam_oidc.so` — and now matches
  `configs/pam/ssh`.
- **`configs/pam/su` and `configs/pam/sudo`** claimed in their headers to maintain
  "traditional authentication for emergency access". They do not: the `requisite
  pam_deny.so` in both refuses anything OIDC did not accept. A comment that misstates
  the stack's semantics is part of the defect, so both now say the stack is OIDC-only
  and that every `su`/`sudo` will wait out the device flow.
- **`README.md`**: the platform table claimed Console and GUI as **Stable** on five
  distributions. CI exercises `sshd` on one image (Debian 12, `test/e2e`), so the table
  now marks that one cell as tested, the console/`su`/`sudo` stacks as untested
  examples, and display managers as unsupported — no `gdm`/`sddm`/`lightdm` stack is
  shipped. The "Subsequent access uses cached SSH key" line is gone: `pam_oidc.so` sends
  no session ID with its request, so the broker has nothing to match a login against and
  every login runs a fresh device flow. The headline and the feature bullet no longer
  claim GUI logins.

## Tests, and proof they were failing

Four new tests in `cmd/pam-module/configs_test.go`, which read the shipped files rather
than describing them. `TestShippedPAMStacksDoNotEnableDebug` (#168) is unchanged in
substance; it now shares the file-enumeration helper, and no stack added here passes
`debug`.

Against the **pre-change** tree (only `configs_test.go` added):

```
--- FAIL: TestShippedPAMStacksAreServiceScoped (0.00s)
    configs_test.go:184: ../../configs/pam/common-auth is the Debian/Ubuntu stack that every service on the host @includes: shipping it puts an interactive device flow in the auth stack of su, sudo, the display manager, polkit and cron. Ship a per-service file instead (one of login, ssh, su, sudo).
--- FAIL: TestShippedPAMStacksUseDocumentedControlFlags (0.00s)
    configs_test.go:258: ../../configs/pam/common-auth:12 controls pam_oidc.so with [success=2 default=ignore]; the shipped stacks use the plain flags (sufficient or required for auth) so that adding a line cannot change what a skip count means: auth    [success=2 default=ignore]     pam_oidc.so
--- FAIL: TestInstallScriptsDoNotTouchAggregatePAMStacks (0.00s)
    configs_test.go:304: ../../scripts/install.sh:208 touches the Debian/Ubuntu host-wide PAM stack "common-auth": an installer must only edit the /etc/pam.d file of the service the operator asked for, never the one every service @includes: sed -i '2i@include common-auth' /etc/pam.d/sshd
```

The fourth is the guard against the docs and the files drifting apart, so it fails on the
intermediate state — stack deleted, docs not yet updated:

```
--- FAIL: TestDocsReferenceOnlyShippedPAMStacks (0.00s)
    configs_test.go:350: ../../configs/pam/README.md:43 refers to configs/pam/common-auth, which this repository does not ship: sudo cp configs/pam/common-auth /etc/pam.d/common-auth
```

All five pass on this branch.

## Verification

- `gofmt -l pkg internal cmd` — empty; `gofmt -s -l .` — empty
- `go build ./...` — ok
- `go test ./pkg/... ./internal/...` — all ok
- `cmd/pam-module` is cgo and does not build on the macOS host this was written on, so
  the container gate:
  `docker run --rm -v "$PWD":/src -w /src -e GOFLAGS=-buildvcs=false golang:1.25 bash -c 'apt-get update -qq && apt-get install -y -qq libpam0g-dev libjson-c-dev && go vet ./cmd/pam-module && go test ./cmd/pam-module/...'`
  → vet clean, `ok github.com/scttfrdmn/oidc-pam/cmd/pam-module 3.030s`
- `golangci-lint` v2.12.2 (the pin in `.golangci-version`) `run ./...` in the same
  container → `0 issues.`

## Deliberately out of scope

- **The module does not yet return `PAM_IGNORE` promptly for a service that cannot do a
  device flow** (fix item 2 and acceptance criterion 2 in #172). Both plausible gates
  are risky in the direction that matters here: a service-name denylist is guesswork, and
  a TTY heuristic that guesses wrong turns a working login into `PAM_IGNORE` → `pam_deny`
  → lockout. The one provable version — bail out when `pam_get_item(PAM_CONV)` yields no
  conversation function, since the verification URL then has nowhere to go — would invert
  `TestAuthenticateSurvivesAMissingConversationFunction`, which #168 wrote deliberately to
  assert the module carries on past an undisplayable message. That is a design decision
  for its own PR with its own e2e case; this PR removes the configuration that made every
  service on the host hit the problem.
- **`PAM_PROMPT_ECHO_ON` instead of `PAM_TEXT_INFO`** (fix item 3) — a change to the
  login UX and to what the e2e harness drives.
- **`scripts/install.sh` still wires PAM unconditionally**, unlike
  `scripts/install-release.sh`, which requires `--configure-pam`. Worth aligning, but it
  is a separate CLI change; what it inserts is now non-fatal either way.


scttfrdmn:
Reviewed and merging. Confirmed against the tree rather than the description:

- `test/e2e` really is Debian bookworm (`Dockerfile.{broker,client}`) and `test/e2e/pam-sshd` really carries the `configs/pam/ssh` auth stack shape, so the README's "Tested in CI" cell is accurate rather than aspirational.
- The "no cached credential" paragraph checks out: `send_auth_request` sends no `session_id`, and `Broker.Authenticate` reuses a session only via `getAndRemoveIfExpiredSession(req.SessionID, req.UserID)` (`pkg/auth/broker.go:307`), so every login does run a fresh flow.
- `docs/design/README.md` exists, so `TestDocsReferenceOnlyShippedPAMStacks` doesn't `t.Fatalf` on a missing file. All five config tests pass locally, `gofmt` clean, and all 17 checks are green including the cgo and e2e jobs.

You were right to flag the `Fixes #172` problem: acceptance criterion 2 (`PAM_IGNORE` for a non-interactive service) and fix item 3 (`PAM_PROMPT_ECHO_ON`) are untouched by this PR, and merging would close the issue over them. Filed as #193, which also notes that the `PAM_TEXT_INFO` question belongs to `oauth2-pam` under #179 rather than being decided here.

Issue #169, opened 2026-08-14
source_ip is never populated by any client, and empty means 'deny': require_private_network denies every login and every user scores 70 risk from their second login on


scttfrdmn:
## What's wrong

`source_ip` is the principal input to Tailscale enforcement, private-network
enforcement, IP allowlists, unusual-location risk and location history. **Neither
client ever populates it.**

- The C module sends `type`, `user_id`, `login_type`, `target_host` and
  `metadata{service,tty,pid}` — `cmd/pam-module/cgo_bridge_linux.c:133-178`. No
  `source_ip` appears anywhere in the C at all.
- `pkg/pam/pam.go:160-201` sends the same set.
- `internal/ipc/validate.go:117-127` then blesses the gap: "source IP is optional",
  empty accepted.

Downstream, empty does not mean *unknown* — it means *bad*.

## Impact

**1. Two config options deny every login on earth.** `pkg/auth/policy.go:154-176`:

```go
if netReq.RequireTailscale  { if !pe.isTailscaleIP(req.SourceIP) { result.Allowed = false; … } }
if netReq.RequirePrivateNetwork { if !pe.isPrivateIP(req.SourceIP) { result.Allowed = false; … } }
```

`isPrivateIP("")` is false. `configs/production/broker-enterprise.yaml:160-163` sets
**both**.

**2. Every user's risk score is 70 from their second login onward.**
`calculateRiskScore` (`policy.go:426-455`): `+15` for an empty `DeviceID` (also never
sent), `+25` for `!isPrivateIP("")`, and `+30` for "unusual location" — because
`LocationHistory.IsUnusual` (`pkg/auth/location_history.go:73-99`) returns `false`
only while the history is empty; once a first entry with `Subnet: ""`/`Country: ""`
exists, both comparisons are guarded by `subnet != ""` / `country != ""`, so they are
skipped and the function falls through to `return true`. So: **40 on the first login,
70 on every login after, 90 after 18:00.** Any `DENY` threshold an operator would
plausibly write locks the whole site out starting on day two.

**3. Resource policies match the wrong end of the connection.**
`matchesResourcePolicy(req.TargetHost, policyName)` (`policy.go:478-485`) — and
`TargetHost` is filled from PAM's `PAM_RHOST`, the *client's* address. See #158.

**4. The broker faithfully records an always-empty IP forever.**
`broker.go:1032` — `b.policyEngine.RecordLocation(updated.UserID, updated.SourceIP)`.

[...]

Issue #165, opened 2026-08-14
RemoveExpiredKeys drops the trailing newline, making the next login's key permanently unrevocable while the audit log reports it as revoked


scttfrdmn:
## What's wrong

`RemoveExpiredKeys` rewrites `authorized_keys` **without a trailing newline**, and
`AddPublicKey` then appends its comment directly onto the last surviving key line.
The resulting line can never be removed again, and the failed removal is audited as a
success.

- `pkg/ssh/authorized_keys.go:315-321` — `bufio.Scanner`, which strips `\n`
- `:360` — `newContent := strings.Join(filteredLines, "\n")` — no trailing newline
- versus `:195-201` — `AddPublicKey` appends `"# Added by OIDC PAM on ...\n"` with
  `O_APPEND`, unconditionally
- and versus `RemovePublicKey` at `:238`/`:261`, which uses `strings.Split(…, "\n")`
  — preserving the trailing empty element, so its `Join` **does** restore the newline.
  The two functions disagree and only the `Scanner` one is wrong.

## Sequence — no attacker required, ordinary operation

1. A sweep removes at least one expired key, so the file loses its final newline.
2. The user's next login appends its comment onto the last surviving line:
   `ssh-rsa AAA... alice@oidc-pam-1770000000# Added by OIDC PAM on 2026-08-13 12:00:00`
3. That line is now permanently unrevocable:
   - `RemoveExpiredKeys` takes `parts[2]` and `strconv.ParseInt` fails on
     `1770000000# Added...`, so the key is **retained** (`:333-346`, deliberately
     fail-safe).
   - `RemovePublicKey` compares the full trimmed line for equality (`:246`), so
     targeted revocation no longer matches either.
4. `revokeSSHKey` treats the no-op as success: `RemovePublicKey` returns `nil` after
   only a `log.Debug()` (`:253-257`), then `broker.go:1274-1281` logs
   `"SSH key revoked"` at Info and records a success metric.

So the audit trail asserts a revocation that did not happen, and the `authorized_keys`
entry survives indefinitely.

## Fix

1. `RemoveExpiredKeys`: `strings.Join(filteredLines, "\n") + "\n"` when non-empty.
2. Defensively, have `AddPublicKey` stat the file and prepend `"\n"` if it does not
   end in one.
3. Make `revokeSSHKey` treat "line not found" as a distinct, audited outcome — never
   log `"SSH key revoked"` for a removal that matched nothing.

## Acceptance criteria

[...]

PR #194, opened 2026-08-14
fix(ssh): resolve homes from the account database and expire issued keys

### LINKED ISSUE #171: SSH key provisioning cannot work as shipped: systemd blocks both writes, homes are guessed, keys accumulate one per login forever, and no case logs in twice
## What's wrong

`README.md:13` promises "Generate, rotate, and revoke SSH keys automatically" and
`README.md:165` promises "Subsequent access uses cached SSH key". Four independent
facts stop that working. **Decision taken: keep the feature and fix it properly** —
this issue is the fix list.

### 1. The shipped systemd unit makes it impossible

`configs/systemd/oidc-auth-broker.service`:

```ini
User=root
ProtectSystem=strict
ProtectHome=true
ReadWritePaths=/var/run/oidc-auth /var/log/oidc-auth /etc/oidc-auth
```

- `ProtectHome=true` makes `/home` inaccessible, so `AddPublicKey` cannot write
  `authorized_keys`.
- The key store is `/var/lib/oidc-pam/ssh-keys` (`pkg/auth/broker.go:216`), absent from
  `ReadWritePaths` under `ProtectSystem=strict`, so `SaveKey` cannot write the keypair.

Both halves must fail on every production install. `DEPLOYMENT.md:437-438` repeats
`ProtectHome=true` and lists `/var/lib/oidc-auth` — not even the real path. Neither
`scripts/install.sh` nor `install-release.sh` creates `/var/lib/oidc-pam`; both create
a service user homed at `/var/lib/oidc-auth` (`install.sh:122`) that the unit does not
use.

**And the failure is silent**: per #152's post-mortem (`test/e2e/README.md:30-32`) a
key-store failure is non-fatal — "every save failed, and the login succeeded anyway
with no key installed". So in production, logins work and keys never appear, with one
log line.

### 2. The home directory is guessed

`pkg/auth/broker.go:217` hardcodes `NewAuthorizedKeysManager("/home")` and
`pkg/ssh/authorized_keys.go` joins `<baseDir>/<username>/.ssh`. No `getpwnam`. On any
site with LDAP/SSSD homes, `/home/<domain>/<user>`, autofs or NFS — i.e. exactly the
research-computing sites the README targets — this writes to a path sshd never reads.
Worse, `MkdirAll(sshDir, 0700)` (`authorized_keys.go:55`) will create
`/home/<user>` itself as root-owned 0700 if absent, locking the user out of their own
home.

### 3. Keys accumulate without bound and nothing can revoke them

`AddPublicKey` (`:150-265`) dedupes only on an **exact identical line**, and
`GenerateKey` embeds a fresh timestamp in the comment (`<user>@oidc-pam-<unix>`), so
every login produces a line that is by construction not a duplicate: **one new
RSA-4096 line per login, appended forever.** The line carries no `expiry-time=` and no
`from=`, so sshd honours it until something physically removes it.

The only remover is `RemoveExpiredKeys`, whose cutoff is a hardcoded 24 hours parsed
from the comment — *not* the configured `token_lifetime` — reached only via
`sweepExpiredAuthorizedKeys` (`broker.go:1166-1188`) for users who happen to have a
session expiring in that pass. Sessions live only in the broker's memory, so a restart
orphans every key ever issued (the function's own comment admits this). There is no
logout hook and no revocation path from the IdP: disable a user in Okta and their key
keeps working for up to 24 hours, or indefinitely after a broker restart.

`pkg/ssh/keys.go:31-35` already carries a `TODO(L-3)` calling expiry-in-a-comment
fragile. It is worse than fragile: it is the only expiry that exists.

The project whose stated problem is *"Key Sprawl: thousands of orphaned keys"* and
*"No Rotation"* (`README.md:24-25`) generates orphaned, unrotatable keys at one per
login per host.

### 4. Nothing consumes the private key

`SaveKey` writes it to `/var/lib/oidc-pam/ssh-keys/<sessionID>/id_rsa` mode 0600 owned
by root. The IPC `Response` (`internal/ipc/server.go:54-72`) carries `SSHPublicKey` and
nothing else; `grep -rn 'private_key\|PrivateKey'` across `internal/ipc`, `broker.go`,
the C bridge and `internal/brokerclient` returns **nothing**. So the installed
`authorized_keys` entry authorizes a keypair whose private half the user cannot read —
today the feature is a side effect nobody consumes. Fixing 1–3 does not by itself make
the key usable as a credential; that is the separate SSH-CA question.

## Why the harness didn't catch it

`expect_oidc_key_for` (`test/e2e/cases.sh:221-235`) asserts a line matching
`@oidc-pam-` exists and that the count is exactly 1 — but **no case logs in twice**, so
the accumulation is invisible. A second `attempt_login alice` in `case_approved_login`
turns this from invisible to a red build.

## Fix

1. Fix the unit: drop `ProtectHome=true` (or set `ProtectHome=read-only` plus explicit
   `ReadWritePaths`), add `/var/lib/oidc-pam`, and prefer `StateDirectory=oidc-pam` so
   systemd creates it with the right ownership. Fix `DEPLOYMENT.md` to match, and make
   the installers create the real path.
2. Resolve homes via `user.Lookup`; refuse to create a missing home directory; `fstat`
   the opened fd and require `st_uid == target uid`.
3. Make a key-store write failure **fail the login** rather than being logged and
   ignored — a silent success that provisions nothing is what hid #152.
4. Dedupe per user rather than per timestamp: one live `@oidc-pam-` key per user, with
   the new key replacing the old. Add `expiry-time=` (and consider `from=`) to the line
   so sshd enforces expiry itself. Drive the cutoff from the configured
   `token_lifetime`, not a hardcoded 24 h.
5. Reconcile on startup: sweep every `@oidc-pam-` key with no live session, so a broker
   restart stops orphaning credentials.
6. Add the second-login e2e assertion.

## Acceptance criteria

- A test asserting two consecutive logins leave exactly **one** `@oidc-pam-` key.
- A test asserting the line carries `expiry-time=` derived from `token_lifetime`.
- A test asserting a key-store write failure denies the login.
- An e2e case: restart the broker mid-session, then assert the orphaned key is swept.
- The systemd unit is exercised (or at least the paths asserted against the unit file)
  so `ProtectHome`/`ReadWritePaths` drift cannot silently disable the feature again.

scttfrdmn:
Fixes #171

## What was broken

The SSH key a successful login was granted was written to a directory guessed from the login name, was never removed when the session ended, and accumulated one working credential per login.

The broker joined `/home` with the login name and wrote `<that>/.ssh/authorized_keys`. On any site whose homes are not literally `/home/<user>` — SSSD/LDAP with `/home/<domain>/<user>`, autofs, NFS, a home moved by hand — it created that directory itself, as root and mode 0700, wrote a key list sshd would never read, and reported the login as successful. The user was told they were authenticated and then could not log in; on their next real login their own home was shadowed by a root-owned directory they could not write. The owning uid from the passwd database was never consulted, so nothing distinguished "this is the account's home" from "this is somebody else's directory".

Where the key *did* land it was appended on every login and deduplicated only against byte-identical lines, so an account that logged in daily accumulated one live credential per login. None of them carried an expiry sshd could see, so a key stopped being limited at all the moment the broker forgot the session that owned it — which happens on every restart, since sessions are held only in memory. And the shipped systemd unit set `ProtectHome=true`, which makes `/home` an empty tmpfs to the service, so on a host installed from it no key could be written at all.

Per the issue decision, the feature is kept and fixed; nothing is stubbed out.

## What changed

- **Homes come from the account database.** New `pkg/ssh/accounts.go`: `LookupAccount` asks `os/user`, then falls back to `getent passwd`. The fallback is not redundancy — the released broker is built `CGO_ENABLED=0`, and without cgo `os/user` parses `/etc/passwd` and never asks NSS, so every SSSD/LDAP account is invisible to it. `getent` asks NSS the way sshd and `login` do, bounded by `context.WithTimeout` **and** `cmd.WaitDelay` (the context alone does not bound it: `Output` waits on the pipe, and a killed `getent` can leave a forked child holding it). Nothing derives a home path any more, and a home that does not exist is refused rather than created.
- **Ownership is checked** on the home, on `.ssh`, and on `authorized_keys` through the open descriptor (`fstat`, not a second path lookup). The account or root is accepted — what sshd's own `StrictModes` accepts — anything else is refused rather than written through. Files the root broker creates in a home are handed to the account.
- **One live key per account, with an expiry sshd enforces.** A login drops every previous `@oidc-pam-` entry as it writes, leaves the account's own keys alone, and installs an entry carrying `expiry-time="…"`, so sshd enforces the lifetime whether or not the broker is running or remembers the session. That option needs **OpenSSH 7.7 or newer**, and the timespec is written in the host's own time zone because the UTC `Z` spelling is 9.1+ (see the review round below). Entry parsing is now options-aware (`pkg/ssh/entry.go`, quote-aware field split), so the sweep and removal recognise their own entries once they carry options.
- **The configured `token_lifetime` reaches both the key store and the sweep.** Both were hardcoded to 24h, so a site that set `token_lifetime: 1h` still handed out day-long keys.
- **Startup revokes what the previous run left behind.** The key store survives a restart even though sessions do not, so anything in it at startup belongs to no live session; `Start` reconciles it.
- **A login whose key could not be provisioned is denied,** not completed. It used to be activated and audited `authentication_successful` with no usable key anywhere. The failure is audited as `ssh_key_provisioning_failed`, the metric is recorded, and the session is dropped.
- **A superseded key is not a failed revocation.** Revoking a key a later login had already replaced no longer emits `ssh_key_revocation_incomplete`; that event means a credential is still live (#165), and firing it for the ordinary two-logins case would have trained operators to ignore it.
- **Packaging:** `configs/systemd/oidc-auth-broker.service` sets `ProtectHome=false` (with the reason inline), lists `/home` in `ReadWritePaths` (`ProtectSystem=strict` otherwise makes it read-only) and declares `StateDirectory=oidc-pam` / `StateDirectoryMode=0700`; `scripts/install-release.sh` creates `/var/lib/oidc-pam/{ssh-keys,locks}` at 0700; DEPLOYMENT.md's unit snippet now matches the shipped one and gained troubleshooting for the three ways provisioning fails.

## Wire contract

**No change to the broker↔module wire protocol,** so no decision is needed from `scttfrdmn/oauth2-pam` (which owns `docs/wire-protocol.md`). Denying a login whose key could not be provisioned uses the existing convention: the broker calls `removeSession`, and the module's next check gets the already-specified `SESSION_NOT_FOUND`. No new field, error code or changed meaning, and no local divergence left unregistered.

## Which test proves which claim

Every one of these was run **both ways**: passing on this branch, and with the corresponding defect mechanically reintroduced in the production file (a scripted patch per row, reverted afterwards). All 18 rows failed as required with the defect present — I ran the whole matrix twice, once before rebasing onto `origin/main` and again after.

| Claim | Test | Defect reintroduced to prove it |
|---|---|---|
| The home comes from the account database, not a base directory | `pkg/ssh TestPathsComeFromTheAccountDatabaseNotABaseDirectory` | `account.Home = filepath.Join("/home", username)` |
| A missing home is refused, never created as root 0700 | `pkg/ssh TestAddPublicKeyRefusesAMissingHomeAndDoesNotCreateIt` | `os.MkdirAll(account.Home, 0700)` on `IsNotExist` |
| A foreign-owned home/.ssh/authorized_keys is refused | `pkg/ssh TestForeignOwnershipIsRefused` | `checkOwner` returns nil |
| A second login supersedes the first key instead of accumulating | `pkg/ssh TestASecondLoginSupersedesTheFirstKey` | match only byte-identical entries |
| The installed entry carries the expiry sshd enforces | `pkg/ssh TestInstalledEntryCarriesTheExpirySSHDEnforces` | drop the `expiry-time=` option |
| The sweep measures against the configured lifetime | `pkg/ssh TestSweepUsesTheConfiguredLifetimeForOlderEntries` | use `DefaultKeyLifetime` |
| Entries carrying options are understood by sweep and removal | `pkg/ssh TestSweepReadsEntriesThatCarryOptions`, `TestRemovalMatchesAnEntryThatCarriesOptions` | `strings.Fields` instead of the options-aware split |
| An account only NSS knows about resolves | `pkg/ssh TestAnAccountKnownOnlyToNSSResolvesThroughGetent` | skip the `getent` fallback |
| A wedged account database cannot hold a login open | `pkg/ssh TestAWedgedAccountDatabaseDoesNotHangTheLookup` | remove `cmd.WaitDelay` (this one found a real bug in my own first cut) |
| A login whose key cannot be provisioned is denied and audited | `pkg/auth TestLoginIsDeniedWhenTheKeyCannotBeProvisioned`, `TestLoginIsDeniedWhenTheKeyStoreCannotBeWritten` | activate the session anyway |
| Startup revokes the previous run's keys | `pkg/auth TestStartupRevokesKeysLeftByAPreviousRun` (drives the real `Start`/`Stop`) | drop the `reconcileIssuedKeys` call |
| `token_lifetime` reaches the key manager | `pkg/auth TestConfiguredTokenLifetimeLimitsTheIssuedKey` (real `NewBroker`) | drop `SetExpiration`/`SetKeyLifetime` |
| A superseded revocation is not reported as incomplete | `pkg/auth TestRevokingASupersededKeyIsNotReportedAsIncomplete` | never consult `KeyIsAuthorized` |
| The shipped unit lets the broker write where it must | `cmd/broker TestShippedUnitLetsTheBrokerWriteWhereItMust` (reads the shipped file) | `ProtectHome=true`, drop `/home` and `StateDirectory` |
| The timespec is the form every sshd that knows the option parses | `pkg/ssh TestTheExpiryTimespecIsTheFormEverySupportedSSHDParses`, `TestInstalledEntryCarriesTheExpirySSHDEnforces` | write the UTC `Z` form |
| The docs state the minimum OpenSSH version, and why | `pkg/ssh TestTheDocsStateTheMinimumOpenSSHVersion` | drop the requirement from DEPLOYMENT.md / README.md (proven separately for each) |
| The docs name the platforms too old to run this | `pkg/ssh TestTheDocsNameThePlatformsThatAreTooOld` | remove the 7.4p1 row from the README table |
| A host with no `/home` can still start the broker | `cmd/broker TestShippedUnitLetsTheBrokerWriteWhereItMust` | `ReadWritePaths=… /home` without the `-` |

`pkg/ssh/accounts_test.go` also covers an unknown account (`getent` exit 2 → `ErrUnknownAccount`), a missing `getent`, passwd entries the broker refuses to act on (empty home, non-numeric uid/gid), an answer about the wrong account, non-login-name input, and that `HomeRootLookup` creates nothing.

## Verification

Run on the rebased branch (`origin/main` = b120357):

- `gofmt -l pkg internal cmd` — no output.
- `go build ./...` — clean.
- `go test -count=1 ./pkg/... ./internal/... ./cmd/broker/...` — `ok pkg/auth 7.026s`, `ok pkg/config 0.950s`, `ok pkg/metrics 1.740s`, `ok pkg/pam 0.701s`, `ok pkg/security 5.073s`, `ok pkg/ssh 7.642s`, `ok internal/brokerclient 1.568s`, `ok internal/ipc 2.108s`, `ok cmd/broker 1.384s`.
- Container gate — not strictly required (no C or cgo file is touched) but run anyway: `docker run --rm … golang:1.25 … go vet ./cmd/pam-module ./pkg/pam && go test ./cmd/pam-module/... ./pkg/pam/...` → exit 0, `ok cmd/pam-module 3.029s`, `ok pkg/pam 0.002s`.
- `bash -n` on `test/e2e/cases.sh`, `test/e2e/run-tests.sh`, `scripts/install-release.sh` — clean.
- Defect matrix: 14/14 reintroductions caught.

## e2e additions (not executed here)

`test/e2e/cases.sh` gains `case_second_login_supersedes_key` (two full device flows: still exactly one `@oidc-pam-` entry, a different key blob, the first blob gone) and `case_orphan_keys_swept` (driven in two phases across a broker restart by `run_orphan_case` in `run-tests.sh`), and `case_approved_login` now plants a personal key, asserts it survives provisioning, and asserts the installed entry carries `expiry-time="`. Both new cases are registered in `CASES`. These were **not run on this host** — the suite needs the full docker-compose stack — so they are best-effort until CI runs them; the Go suite is what proves the claims above.

## Deliberately out of scope

- Section 4 of the issue (how the *private* key is consumed, and whether this should be an SSH CA instead of authorized_keys). That is a design change to the login flow, not a defect in provisioning, and wants its own issue.
- Adding a `from=` restriction to the installed entry. Now that #169 makes the real source address available it is worth doing, but it is a policy addition rather than part of this fix.

## Review round 2 (commit `ca0e745`)

Two findings from review, plus one the first finding turned up.

1. **The `Z` was worse than the version note.** The reviewer flagged that `expiry-time=` is version-gated and only a Go comment said so. Checking the numbers rather than taking them: `expiry-time=` was added in **OpenSSH 7.7** (7.7 release notes, April 2018) — so RHEL 8's 8.0p1 does know the option — but the **`Z` suffix meaning UTC was only added in 9.1** (9.1 release notes). Every sshd before 9.1 parses a timespec by its length (8, 12 or 14 digits) and rejects the 15-character `Z` form, taking the whole entry with it. So the first commit worked on Debian 12 (9.2) — the one image `test/e2e` runs — and broke RHEL 8 *and 9*, Debian 11, and Ubuntu 20.04 and 22.04. The timespec is now the host-local 14-digit form; the parser still reads the UTC form, because the sweep has to agree with a 9.1+ sshd about entries it did not write.
2. **The requirement is documented.** DEPLOYMENT.md prerequisites state OpenSSH 7.7+ with the reason, the check (`ssh -V`), and the versions the common platforms ship; the supported-OS list drops Amazon Linux 2 and names RHEL/CentOS 7 as excluded (both `7.4p1`). README.md's platform table gains an OpenSSH column and a `❌ Not supported` row for those platforms, and its "expected to work" definition now includes "the platform's sshd is new enough for the `expiry-time=` option" — so #192's honesty is preserved rather than quietly falsified. Versions verified from primary sources: Debian/Ubuntu package pages, Red Hat's RHEL 9 release notes (8.7p1, and 8.0p1 in RHEL 8), Fedora's `openssh-8.7p1-2.fc35`, and the AL2 advisories (`openssh-7.4p1-22.amzn2`).
3. **`ReadWritePaths=-/home`.** systemd fails the unit's namespace setup when a listed path is missing, so the site told to add `/export/home` — the one with no `/home` — would have got a broker that refuses to start. The `-` prefix makes it non-fatal; the comment says why, and the test now requires the prefix rather than merely the path.

No sshd version detection, no config toggle: filed as **#199** with the three options (detect and omit, refuse to start, or make it configurable) and the caveat that a version probe is a heuristic, since the sshd binary found is not necessarily the sshd that will authenticate the session. Both documents link to it.

Verification re-run on `ca0e745`, rebased on `origin/main` (`b120357`): `gofmt -l pkg internal cmd` empty; `go build ./...` clean; `go test -count=1 ./pkg/... ./internal/... ./cmd/broker/...` all `ok`; the `pkg/ssh` suite also run under `TZ=America/New_York` and `TZ=Asia/Kolkata`, since the timespec is now local time; container gate exit 0 again (`ok cmd/pam-module`, `ok pkg/pam`). Defect matrix now **19/19** — the original 14 plus the four new rows above (the docs row proven once per document).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,80 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Security
+- **High (#171): the SSH key a successful login was granted was written to a
+  directory guessed from the login name, was never removed when the session ended,
+  and accumulated one working credential per login.** The provisioning path was
+  wrong at both ends: the key often did not arrive where sshd reads it, and when it
+  did arrive nothing took it away again.
+
+  The broker joined `/home` with the login name and wrote `<that>/.ssh/authorized_keys`.
+  On any site whose homes are not literally `/home/<user>` — SSSD or LDAP with
+  `/home/<domain>/<user>`, autofs, NFS, a home moved by hand — it created that
+  directory itself, as root and mode 0700, wrote a key list sshd would never read,
+  and reported the login as successful. The user was told they were authenticated
+  and then could not log in, and on the next real login their own home was shadowed
+  by a root-owned directory they could not write. The uid from the passwd database
+  was never consulted, so nothing distinguished "this is the account's home" from
+  "this is a directory belonging to someone else". Where the key *did* land, it was
+  appended on every login and deduplicated only against byte-identical lines, so an
+  account that logged in daily accumulated a live credential per login; each one
+  carried no expiry sshd could see, so it stopped being limited at all the moment
+  the broker forgot the session that owned it — which happens on every restart,
+  since sessions are held only in memory. The shipped systemd unit set
+  `ProtectHome=true`, which makes `/home` an empty tmpfs to the service, so on a
+  host installed from it no key could be written at all.
+
+  - Home directories now come from the account database, through
+    `pkg/ssh.LookupAccount`: `os/user` first, then `getent passwd`. The fallback is
+    not redundancy — the released broker is built `CGO_ENABLED=0`, and without cgo
+    `os/user` parses `/etc/passwd` and never asks NSS, so every SSSD or LDAP account
+    is invisible to it. `getent` asks NSS the way sshd and `login` do, bounded by a
+    timeout and a `WaitDelay` so a wedged directory server cannot hold a login open.
+    Nothing derives a home path any more, and a home that does not exist is refused
+    rather than created.
+  - The owning uid is checked, on the home directory, on `.ssh` and on
+    `authorized_keys` through the open file descriptor. The account or root is
+    accepted, which is what sshd's own `StrictModes` accepts; anything else is
+    refused rather than written through. Files the root broker creates in a home are
+    handed to the account.
+  - A login installs exactly one broker key: every previous `@oidc-pam-` entry is
+    dropped as it is written, and the account's own keys are left alone. The entry
+    carries `expiry-time="…Z"`, so **sshd** enforces the key's lifetime whether or
+    not the broker is running or remembers the session. Both the key store and the
+    `authorized_keys` sweep now use the configured `token_lifetime`; both used to be
+    hardcoded to 24 hours, so a site that had deliberately set `token_lifetime: 1h`
+    still handed out keys good for a day.
+  - Startup revokes what the previous run left behind. The key store survives a
+    restart even though sessions do not, so anything in it at startup belongs to no
+    live session and its `authorized_keys` entry is removed.
+  - A login whose key could not be provisioned is now denied instead of completed.
+    It used to be activated and audited as `authentication_successful` with no usable
+    key anywhere; the failure is recorded as `ssh_key_provisioning_failed` and the
+    session is dropped, which the module already reports through
+    `SESSION_NOT_FOUND` — no change to the broker↔module wire contract.
+  - Revoking a key a later login had already superseded is no longer reported as
+    `ssh_key_revocation_incomplete`. That event means a credential is still live
+    (#165), and firing it for the ordinary two-logins case would have trained
+    operators to ignore it.
+  - `configs/systemd/oidc-auth-broker.service` sets `ProtectHome=false`, lists
+    `/home` in `ReadWritePaths` (`ProtectSystem=strict` otherwise makes it
+    read-only) and declares `StateDirectory=oidc-pam`; `scripts/install-release.sh`
+    creates `/var/lib/oidc-pam/{ssh-keys,locks}` at 0700, and DEPLOYMENT.md's unit
+    matches the shipped one.
+
+  Coverage: `pkg/ssh` tests that a key is written to the home the account database
+  gives and that nothing is created at `/home/<user>`, that a missing home is
+  refused rather than created, that a foreign-owned path is refused, that a second
+  login supersedes the first key while leaving the user's own, that the installed
+  entry carries the `expiry-time` option, and that the sweep measures against the
+  configured lifetime; `pkg/ssh/accounts_test.go` covers the NSS fallback, an
+  unknown account, a wedged account database and the passwd fields that are refused;
+  `pkg/auth` tests that a login whose key cannot be installed is denied and audited,
+  that startup revokes a previous run's keys through `Start` itself, that
+  `token_lifetime` reaches the key manager, and that a superseded revocation is not
+  reported as incomplete; `cmd/broker` pins the unit's `ProtectHome`,
+  `ReadWritePaths` and `StateDirectory`; and two e2e cases cover a second login
+  replacing the first key and a broker restart sweeping the keys it had issued.
 - **Medium (#188): stopping the audit logger discarded every record still
   queued, and a second `Stop` panicked.** `processEvents` selected on its stop
   channel and returned without looking at the buffer, so the events sitting in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,11 +46,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     handed to the account.
   - A login installs exactly one broker key: every previous `@oidc-pam-` entry is
     dropped as it is written, and the account's own keys are left alone. The entry
-    carries `expiry-time="…Z"`, so **sshd** enforces the key's lifetime whether or
+    carries `expiry-time="…"`, so **sshd** enforces the key's lifetime whether or
     not the broker is running or remembers the session. Both the key store and the
     `authorized_keys` sweep now use the configured `token_lifetime`; both used to be
     hardcoded to 24 hours, so a site that had deliberately set `token_lifetime: 1h`
     still handed out keys good for a day.
+  - **That option makes OpenSSH 7.7 or newer a requirement**, and it is now stated
+    where an operator will see it: DEPLOYMENT.md's prerequisites, and README.md's
+    platform table, which gains an OpenSSH column and marks Amazon Linux 2 and
+    RHEL/CentOS 7 (both `7.4p1`) unsupported rather than "expected to work". sshd
+    refuses an entire authorized_keys entry that carries an option it does not
+    recognise, so on an older host every key the broker installs is rejected — the
+    same "authenticated, then `Permission denied (publickey)`" outcome, reached a
+    different way. The timespec is written in the host's own time zone, not UTC: the
+    `Z` suffix meaning UTC is OpenSSH 9.1 and newer, and every sshd before that
+    rejects the entry carrying it, which would have limited this to Debian 12-era
+    hosts while breaking RHEL 8 and 9, Debian 11 and Ubuntu 20.04 and 22.04. The
+    broker does not detect the local sshd version; whether it should, refuse to
+    start, or make the option configurable is #199.
   - Startup revokes what the previous run left behind. The key store survives a
     restart even though sessions do not, so anything in it at startup belongs to no
     live session and its `authorized_keys` entry is removed.
@@ -64,10 +77,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     (#165), and firing it for the ordinary two-logins case would have trained
     operators to ignore it.
   - `configs/systemd/oidc-auth-broker.service` sets `ProtectHome=false`, lists
-    `/home` in `ReadWritePaths` (`ProtectSystem=strict` otherwise makes it
+    `-/home` in `ReadWritePaths` (`ProtectSystem=strict` otherwise makes it
     read-only) and declares `StateDirectory=oidc-pam`; `scripts/install-release.sh`
     creates `/var/lib/oidc-pam/{ssh-keys,locks}` at 0700, and DEPLOYMENT.md's unit
-    matches the shipped one.
+    matches the shipped one. The `-` prefix is load-bearing: systemd fails a unit
+    whose `ReadWritePaths` names a path that does not exist, and the site told to add
+    its own home path — `/export/home` and the like — is exactly the site with no
+    `/home`, which would have got a broker that refuses to start.
 
   Coverage: `pkg/ssh` tests that a key is written to the home the account database
   gives and that nothing is created at `/home/<user>`, that a missing home is
@@ -79,9 +95,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `pkg/auth` tests that a login whose key cannot be installed is denied and audited,
   that startup revokes a previous run's keys through `Start` itself, that
   `token_lifetime` reaches the key manager, and that a superseded revocation is not
-  reported as incomplete; `cmd/broker` pins the unit's `ProtectHome`,
-  `ReadWritePaths` and `StateDirectory`; and two e2e cases cover a second login
-  replacing the first key and a broker restart sweeping the keys it had issued.
+  reported as incomplete; `pkg/ssh/docs_test.go` fails if DEPLOYMENT.md or README.md
+  stops stating the minimum OpenSSH version, or if the timespec written is anything
+  other than the form an older sshd parses; `cmd/broker` pins the unit's
+  `ProtectHome`, its optional `-/home` and `StateDirectory`; and two e2e cases cover
+  a second login replacing the first key and a broker restart sweeping the keys it
+  had issued.
 - **Medium (#188): stopping the audit logger discarded every record still
   queued, and a second `Stop` panicked.** `processEvents` selected on its stop
   channel and returned without looking at the buffer, so the events sitting in
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -112,10 +112,41 @@ into your boot/login automation.
 
 #### Minimum Requirements
 - Linux server with PAM support
+- **OpenSSH 7.7 or newer** on every host the broker provisions keys on
 - 2 CPU cores
 - 4 GB RAM
 - 20 GB storage
 - Network connectivity to OIDC provider
+
+##### Why OpenSSH 7.7 is a hard requirement
+
+The broker writes each login's key into `~/.ssh/authorized_keys` with an
+`expiry-time="…"` option in front of it, so that **sshd** stops honouring the key
+once it is stale. That is deliberate: the broker holds its sessions in memory, so
+before this every key it had issued outlived the process that was meant to revoke
+it (#171), and an expiry only the broker knows about is an expiry a restart
+forgets.
+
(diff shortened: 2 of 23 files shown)


scttfrdmn:
Both findings addressed in `ca0e745` (a second commit, not an amend, so the review diff stays readable). Full detail is in the "Review round 2" section of the PR body; the short version:

**Finding 1 turned out to be worse than reported, in a useful way.** Checking the versions rather than taking them: `expiry-time=` was added in **OpenSSH 7.7** (April 2018), so RHEL 8's `8.0p1` does know the option. But the **`Z` suffix meaning UTC is 9.1+** — every sshd before 9.1 parses a timespec by its length and rejects the 15-character form, refusing the whole entry. So the first commit was broken not just on RHEL 8, but on RHEL 8 *and* 9, Debian 11, Ubuntu 20.04 and 22.04 — everything except the Debian 12 image `test/e2e` happens to run. Fixed by writing the host-local 14-digit form, which is what sshd reads an unsuffixed timespec as. The parser still accepts the UTC form, since the sweep must agree with a 9.1+ sshd about entries it did not write.

- DEPLOYMENT.md prerequisites: OpenSSH 7.7+, the reason, `ssh -V`, per-platform versions, and Amazon Linux 2 / RHEL-CentOS 7 (`7.4p1`) called out as unable to run this. The supported-OS list no longer lists AL2 as supported.
- README.md platform table: an OpenSSH column, a `❌ Not supported` row for the `7.4p1` platforms, and "expected to work" now includes "the platform's sshd is new enough for the `expiry-time=` option". #192's table stays true rather than being quietly falsified one commit later.
- Versions from primary sources (Debian/Ubuntu package pages, Red Hat's RHEL 9 release notes, `openssh-8.7p1-2.fc35`, the AL2 advisories), not from memory.
- `pkg/ssh/docs_test.go` pins it: both documents must say "OpenSSH 7.7 or newer" *and* name `expiry-time` as the reason, must name `7.4p1` and Amazon Linux 2, and what `formatExpiryTimespec` writes must be 14 digits with no `Z` that reads back as the same instant. The version lives in one constant, `minOpenSSHVersion`, which the test reads — so code and docs cannot drift.

**Finding 2:** `ReadWritePaths=… -/home`, with the comment explaining that systemd fails the unit when a listed path is missing and that the site told to add `/export/home` is exactly the site with no `/home`. The test now requires the `-` prefix rather than merely the path, so it can't be dropped again.

**No detection or toggle**, as directed: filed as #199 with your three options, plus the caveat that a version probe is a heuristic (the sshd binary found is not necessarily the sshd that authenticates the session). Both documents and the CHANGELOG entry link to it.

Verification on `ca0e745`, rebased on `origin/main` (`b120357`, unchanged since the first push): `gofmt` empty, `go build ./...` clean, full `go test` green, `pkg/ssh` also green under `TZ=America/New_York` and `TZ=Asia/Kolkata` (the timespec is local time now), container gate exit 0. Defect matrix **19/19** — the original 14 plus four new rows, each run with the defect reintroduced and reverted.

Not merged, as asked.

scttfrdmn:
Merging. One correction to my own review, and independent confirmation of yours.

I asserted `expiry-time=` was OpenSSH 8.2+; you checked and said 7.7+, with the real cliff being the `Z` UTC suffix at 9.1+. Rather than settle it from release notes either way, I probed real sshd builds — a generated key, a loopback login, both timespec forms:

```
OpenSSH_8.4p1 Debian-5+deb11u7   utc-Z-suffix: REFUSED    bare-14-digit: ACCEPTED
OpenSSH_9.2p1 Debian-2+deb12u10  utc-Z-suffix: ACCEPTED   bare-14-digit: ACCEPTED
```

That confirms both halves: 8.4 knows the option (or the bare form would have been refused too) and rejects only the `Z` spelling, and the unsuffixed 14-digit form works on both. My premise was wrong, and the defect you found underneath it was worse than the one I reported — the first commit worked on Debian 12, which is the only image `test/e2e` runs, and would have refused every provisioned key on RHEL 8 and 9, Debian 11 and Ubuntu 20.04 and 22.04. Authenticated, then `Permission denied (publickey)`, with the file looking correct. Exactly #171's failure mode reached from the other side, and CI could not have seen it.

Two things I specifically like about how you handled it: `parseExpiryTimespec` reads more forms than `formatExpiryTimespec` writes, so the sweep still agrees with a 9.1+ sshd about entries it did not write; and `minOpenSSHVersion` is one constant that `pkg/ssh/docs_test.go" reads, so the number in the code and the number in DEPLOYMENT.md/README.md cannot drift. Keeping the correction as a second commit rather than an amend was right too — the review diff stays readable.

Also correct not to build the version probe here. #199 carries that decision, and your caveat there is the important part: the sshd binary a probe finds is not necessarily the sshd that authenticates the session.

Issue #204, opened 2026-08-14
HIGH: the .ssh directory is trusted by name, and RemovePublicKey/RemoveExpiredKeys/KeyIsAuthorized/ListOIDCKeys never check it at all


scttfrdmn:
Found by the v0.5.0 security audit. Verified at `d8ca4b3`.

`.ssh` is validated by **name**, never through a held descriptor — and three
functions never validate it at all, so for those there is no race to win.

## Evidence

- `pkg/ssh/authorized_keys.go:368-398` — `userPaths` `Lstat`s the *home*, then
  string-joins `.ssh` and `authorized_keys` (`:396-397`). Nothing is held open.
- `:449-464` — `AddPublicKey` calls `ensureSecureSSHDir` **outside** the lock,
  then takes the lock, then opens by path. The window spans `ensureLockDir` +
  `OpenFile` + `Flock` with a bounded retry of up to 5 s (`:37`, `:86-102`), so a
  second concurrent login for the same account **guarantees** the loser a
  multi-second window after its validation.
- `:525` (`RemovePublicKey`) and `:673` (`RemoveExpiredKeys`) use
  `os.Stat(sshDir)` — which **follows** symlinks — and never call
  `ensureSecureSSHDir`. `KeyIsAuthorized` (`:595`) and `ListOIDCKeys` (`:747`)
  don't check `.ssh` either.
- `ensureSecureSSHDir` itself (`:149-150`) uses `os.Lstat(sshDir)` — correct
  against a symlink *at that instant*, worthless afterwards.
- `syscall.O_NOFOLLOW` at `:202` and `:245` constrains only the **final**
  component; a symlinked `.ssh` is traversed normally.
- `pkg/ssh/accounts.go:202-214` — `checkOwner` accepts `uid == 0`, so a
  **root-owned** redirect target passes the descriptor check.
- `pkg/ssh/security_hardening_test.go:51-70` tests only the case where `.ssh` is
  *already* a symlink when `AddPublicKey` starts.

## Failure scenario

`bob` seeds `~/.ssh/authorized_keys` with lines of his choosing (all non-broker
lines are preserved verbatim, `:474-483`) plus one stale broker entry. He starts
two logins so one parks in `acquireLock` for up to 5 s with `.ssh` already
validated, then:

```sh
mv ~/.ssh ~/.s && ln -s /root/.ssh ~/.ssh
```

The winner reads `/root/.ssh/authorized_keys` — **accepted, because `checkOwner`
[...]

PR #240, opened 2026-08-14
fix(install): make the installers survivable, and put the module where PAM looks (#207, #208, #209, #220)

### LINKED ISSUE #200: CRITICAL: installers hand the broker's socket directory to an unprivileged account, and no client authenticates the broker — unlink+rebind is a full authentication bypass
Found by the v0.5.0 security audit (two independent auditors reached this
independently). Every line below was re-verified by hand at `d8ca4b3`.

## The primitive

Both installers create the broker's socket directory world-readable and then
hand it to an unprivileged account:

- `scripts/install.sh:112-113` — `mkdir -p "$RUN_DIR"; chmod 755 "$RUN_DIR"`
- `scripts/install.sh:136` — `chown -R oidc-auth:oidc-auth "$RUN_DIR"`
- `scripts/install.sh:252-254` — same again
- `scripts/install-release.sh:66` — `install -d -m 0755 ... "$RUN_DIR"`
- `scripts/install-release.sh:78` — `chown -R oidc-auth:oidc-auth "$RUN_DIR" "$LOG_DIR"`

The broker does **not** run as that account — `configs/systemd/oidc-auth-broker.service:9`
is `User=root`. The `oidc-auth` account is a vestigial installer artifact that
nothing runs as.

`internal/ipc/server.go:102-105` does `os.MkdirAll(socketDir, 0750)`, which is a
no-op when the directory already exists, so the installer's `0755 oidc-auth`
survives. Nothing anywhere checks the parent directory's owner or mode.

## Why it is a full bypass

The socket file's own `0660 root` mode is irrelevant: **write permission on the
directory is sufficient to `unlink()` the socket and `bind()` a replacement.**

And no client authenticates the broker. The trust check exists in exactly one
direction — the server verifies its peers (`internal/ipc/peer_creds_linux.go:22-31`,
"peer UID %d is not root") but no client verifies the server:

- `cmd/pam-module/cgo_bridge_linux.c:53-86` — `connect_to_broker` validates the
  path length, connects, and trusts the JSON reply
- `internal/brokerclient/client.go:229` and `cmd/oidc-admin/admin.go:294` — likewise

A repo-wide grep for `SO_PEERCRED`/`getsockopt` across `cmd/pam-module`,
`internal/brokerclient`, `cmd/oidc-admin` and `pkg/pam` returns **nothing**.

## Failure scenario

Anything executing as uid `oidc-auth` — a compromised sidecar, a site-local cron,
a log shipper pointed at the `oidc-auth`-owned `/var/log/oidc-auth` — does:

```
unlink("/var/run/oidc-auth/broker.sock")
bind(own AF_UNIX socket at the same path)
```

The real broker keeps serving on its now-unlinked inode, logs nothing, and
`systemctl status` stays green. Every subsequent PAM login connects to the
attacker's listener, which replies `{"success":true}\n`. `classify_response`
(`cgo_bridge_linux.c:614-651`) sees a boolean `success` true and no
`requires_device`, and returns `PAM_SUCCESS`. With the shipped
`auth sufficient pam_oidc.so` (`configs/pam/ssh:23`, `configs/pam/login:18`,
`configs/pam/su`, `configs/pam/sudo`) that short-circuits the auth stack:
`ssh root@host`, `su -` and `sudo` all succeed with no credential.

uid `oidc-auth` → uid 0. A broker restart evicts the attacker
(`server.go:97` `RemoveAll`), who simply retakes the path.

The same directory ownership lets that uid truncate or symlink
`/var/log/oidc-auth/audit.log`, destroying the trail.

**Reachability, stated honestly:** `oidc-auth` has `-s /bin/false` and no
password, and no shipped component runs as it, so this needs a prior foothold as
that specific uid. Filed as critical anyway because the outcome is
authentication as any user including root, the fix is a few lines, and the
installers create the primitive for **no benefit at all** — the account they
chown to is unused.

## Fix

1. Stop chowning the runtime directory. Replace the installers' `mkdir`/`chown`
   with `RuntimeDirectory=oidc-auth` + `RuntimeDirectoryMode=0750` in the unit so
   systemd creates it root-owned on every start (this also fixes the
   reboot-survival bug filed separately).
2. Have `Server.Start` refuse to bind when `filepath.Dir(socketPath)` is not
   root-owned or is group/other-writable.
3. Defence in depth: have the clients verify the *server*. On Linux
   `SO_PEERCRED` on a connected client socket returns the listener's credentials,
   so the module, `brokerclient` and `oidc-admin` can each require uid 0 —
   exactly what `verifyPeerCredentials` already does in the other direction.

## Test

Yes, cheaply: a Go test that binds a fake broker at a temp socket path and
asserts `brokerclient` refuses a non-root peer; a startup test asserting `Start`
fails when the socket directory is `0777` or owned by another uid; and an e2e
case that binds a rogue socket as a non-root uid and asserts the login is
refused.

### LINKED ISSUE #207: HIGH: uninstall.sh locks every user out of SSH by leaving 'auth requisite pam_deny.so' as the entire auth phase
Found by the v0.5.0 security audit. Verified at `d8ca4b3`.

`scripts/uninstall.sh` locks **every user out of SSH, permanently**, on any host
wired with this project's own documented PAM stack — and prints
"Uninstallation completed successfully!" while doing it.

## Evidence

`scripts/uninstall.sh:79-83`:

```sh
if grep -q "pam_oidc.so" /etc/pam.d/sshd 2>/dev/null; then
    sed -i '/pam_oidc.so/d' /etc/pam.d/sshd
    sed -i '/# OIDC PAM Authentication/d' /etc/pam.d/sshd
```

against `configs/pam/ssh:23-24`:

```
auth    sufficient  pam_oidc.so
auth    requisite   pam_deny.so
```

## Failure scenario

An operator follows `configs/pam/README.md` and installs `configs/pam/ssh` as
`/etc/pam.d/sshd` — the one stack CI exercises end to end. Later they run
`sudo ./scripts/uninstall.sh`.

The `sed` deletes lines 23, 30, 36 and 43 (every line containing `pam_oidc.so`).
What remains as the **entire auth phase** is:

```
auth    requisite   pam_deny.so
```

Every SSH authentication is now refused, for every user, permanently. The script
prints success at `:224`.

## Aggravating factors

- **The rollback is dead code.** `:85-88` `[[ -f ... ]]`-tests the backup the
  script wrote seconds earlier and only calls `print_info`. There is no restore
  path at all — and the `$(date +%Y%m%d-%H%M%S)` in the test can differ by a
  second from the one at `:75`, so even the message is racy.
- **It removes the module while leaving other stacks referencing it.** `:106`
  deletes `/lib/security/pam_oidc.so` but only touches `/etc/pam.d/sshd`, so
  `/etc/pam.d/sudo`, `su` and `login` — if configured per `configs/pam/*` — are
  left pointing at a module that no longer exists, each also backed by
  `requisite pam_deny.so` (`configs/pam/sudo:19-20`, `configs/pam/su:18-20`).
  **`sudo` and `su` then deny everyone, so the operator cannot escalate to repair
  the sshd stack.**
- `:169-170` runs `rm -f /tmp/oidc-auth-* /tmp/pam_oidc-*` as root against a
  world-writable directory. An unprivileged user who pre-creates a *directory* at
  `/tmp/oidc-auth-x` makes `rm -f` exit non-zero, and `set -e` (`:6`) aborts the
  uninstall before `print_summary`.
- It never removes `/var/lib/oidc-pam`, so the SSH private keys the broker issued
  stay on disk after uninstall, and it never sweeps broker-written
  `authorized_keys` entries from user homes. Those are at least bounded, because
  entries carry `expiry-time=` (`pkg/ssh/authorized_keys.go:301`, `:515`).

## Fix

- Delete only the lines the installer added — match on a sentinel comment the
  installer writes.
- After any edit, validate that the resulting stack has at least one
  non-`pam_deny` auth module, and restore the backup if not.
- Extend cleanup to **every** `/etc/pam.d/*` file that references `pam_oidc.so`,
  and refuse to remove the `.so` while any stack still references it.
- `rm -rf --` on explicit paths, or skip the `/tmp` cleanup entirely.
- Sweep broker-issued keys and remove `/var/lib/oidc-pam` (with a prompt).

## Test

Yes, cheaply — the e2e client container already holds a real PAM stack. A case
that copies `configs/pam/ssh` to `/etc/pam.d/sshd`, runs the uninstaller's
`remove_pam_config`, and asserts a **password** login still succeeds would fail
today. Nothing in CI executes either installer.

### LINKED ISSUE #208: HIGH: installers hardcode PAM_DIR=/lib/security, which exists on neither Debian/Ubuntu nor RHEL — install aborts, or the module silently never loads
Found by the v0.5.0 security audit. Verified at `d8ca4b3`, and confirmed
empirically against real distro images.

Both installers hardcode a PAM module directory that exists on **neither**
Debian/Ubuntu nor RHEL/Amazon Linux.

## Evidence

- `scripts/install-release.sh:21` — `PAM_DIR="/lib/security"`, used at `:95`
  (`install -m 0644 "$pam_mod" "$PAM_DIR/pam_oidc.so"`) under `set -euo pipefail`
  (`:18`)
- `scripts/install.sh:10` and `:160` — same
- `Makefile:122`, `:133` — same
- Documented as correct in `DEPLOYMENT.md:245`, `:538-539`,
  `configs/pam/README.md:57`, `:298-301`, `REQUIREMENTS.md:80`, `:85`

Measured:

```
$ docker run --rm debian:bookworm-slim ls -d /lib/security
ls: cannot access '/lib/security': No such file or directory
$ docker run --rm debian:bookworm-slim dpkg -L libpam-modules | grep pam_permit
/lib/aarch64-linux-gnu/security/pam_permit.so
$ docker run --rm amazonlinux:2023 ls -d /lib/security       → No such file or directory
$ docker run --rm amazonlinux:2023 ls -d /usr/lib64/security → /usr/lib64/security
```

**The project's own test harness already knows this.**
`test/e2e/Dockerfile.client:52` derives the directory with
`dpkg -L libpam-modules`, commented *"Ask dpkg where those are rather than
hardcoding a multiarch triplet."* The installer that ships to operators does the
exact thing the harness deliberately avoids.

## Failure scenario A — today, every Debian/Ubuntu and RHEL host

An operator downloads the v0.5.0 tarball and runs the canonical
`sudo ./install.sh` (`release.yml:142` ships `install-release.sh` as the archive
root `install.sh`). `install_binaries` puts three binaries in `/usr/local/bin`,
then `:95` fails with "No such file or directory" and `set -e` aborts. No config,
no unit, no summary — a silently half-installed host. This is the primary
documented install path (`DEPLOYMENT.md:230-249`).

## Failure scenario B — the dangerous one

The operator "fixes" it with `mkdir -p /lib/security` and re-runs. The install now
succeeds, but libpam's compiled-in module path on those distros is
`/lib/<triplet>/security` or `/usr/lib64/security`, so `pam_oidc.so` is never
found:

- on a host wired with `configs/pam/ssh`, the unresolvable `sufficient` module
  falls through to `auth requisite pam_deny.so` → **every SSH login refused**
- on a host wired by `install.sh:218-219`, the module silently never runs, so the
  operator believes OIDC is enforced while password auth still decides —
  **fail-open by misconfiguration**

## Fix

Resolve the directory at install time and hard-fail if none exists, *before*
touching anything else:

```sh
dirname "$(dpkg -L libpam-modules 2>/dev/null | grep -m1 '/pam_permit\.so$')"
# else first existing of: /usr/lib64/security, /usr/lib/$(gcc -dumpmachine)/security, /lib/security
```

Update the four documents.

## Test

A container test that runs `install-release.sh` against `debian:bookworm-slim` and
`amazonlinux:2023` and asserts exit 0 plus `pam_oidc.so` on libpam's actual search
path. Nothing in CI executes either installer today.

## Not verified

Scenario A is directly verified. Scenario B's second-order claim — that libpam's
compiled-in `DEFAULT_MODULE_PATH` has no `/lib/security` fallback — is inference;
libpam was not disassembled. Scenario A stands on its own.

### LINKED ISSUE #209: HIGH: both installers write broker.yaml — the AES-256 token key and every client secret — at mode 0644, and the runtime check only warns
Found by the v0.5.0 security audit. Verified at `d8ca4b3`.

The shipped install path makes the one file holding all long-term key material
world-readable on a multi-user host — which is exactly the host type this product
exists for. It also defeats the fail-closed key validation added in v0.4.0: the
key is validated for *strength*, then published to every local user.

## Evidence

- `scripts/install.sh:177` — `chmod 644 "$CONFIG_DIR/broker.yaml"`
- `scripts/install-release.sh:116` — `install -m 0644 "$example_cfg" "$CONFIG_DIR/broker.yaml"`

What is in that file:
- `configs/examples/broker.yaml:149` — `token_encryption_key: "REPLACE-with-output-of-oidc-admin-gen-key"`
- `configs/production/broker-minimal.yaml:27` — `client_secret: "your-client-secret"`; `:72` — `token_encryption_key:`

**The config's own instructions contradict the installer** —
`configs/production/broker-minimal.yaml:112-114`:

```
# File permissions should be:
# chown root:root /etc/oidc-auth/broker.yaml
# chmod 600 /etc/oidc-auth/broker.yaml
```

And the runtime check is warn-only, with a mask that does not even cover
group-read — `pkg/config/config.go:375-379`:

```go
if info, err := os.Stat(configPath); err == nil {
    if mode := info.Mode().Perm(); mode&0137 != 0 {
        log.Printf("WARNING: config file %s has permissions %04o, ...", configPath, mode)
    }
}
```

`0137` is owner-x, group-wx, other-rwx. It catches other-read, but only *warns*,
on stdlib `log` (stderr) rather than zerolog — so it is easy to miss in
journald-based aggregation — and the broker proceeds to serve.

## Failure scenario

Operator runs `sudo ./install.sh`, follows QUICK-START, pastes
`oidc-admin gen-key` output into `/etc/oidc-auth/broker.yaml:149` and the Okta
client secret at `:27`. Any local unprivileged user runs
`cat /etc/oidc-auth/broker.yaml` and obtains:

1. the **client secret** — usable directly against the IdP token endpoint
2. the **AES-256 key** that is the sole protection on tokens in the broker's heap

The broker logs one `WARNING` line to stderr at startup and runs normally.

## Related, same root cause

- `scripts/install.sh:116` — `chmod 755 "$LOG_DIR"` and `:249`
  `chmod 644 "$LOG_DIR/broker.log"` make the broker's own log world-readable,
  which is the delivery mechanism for the debug-logged device code (filed
  separately).
- `DEPLOYMENT.md:529` tells operators to `chown -R oidc-auth:oidc-auth
  /etc/oidc-auth` while the unit runs the broker as `User=root`
  (`configs/systemd/oidc-auth-broker.service:9`) — so the "Security Hardening"
  section hands config write access to an unprivileged account. That account can
  then set `skip_tls_verify: true`, add a rogue provider, blank `require_groups`,
  or set `audit.enabled: false` — full control of who may authenticate to the
  host. Fix the docs to `chown root:root` + `chmod 600`, and drop
  `/etc/oidc-auth` from `ReadWritePaths` (`:41`) — the broker should not write its
  own config.

## Fix

`install -m 0600 -o root -g root` in both installers, and make `LoadConfig`
**fail** rather than warn when `mode&0177 != 0` (the mask must include
group-read), with `OIDC_AUTH_DEV=true` as the only escape — matching the pattern
already at `pkg/config/config.go:760-790`.

## Test

Two, and there is precedent for both: a test asserting the installers'
`chmod`/`install` mode literal (as `cmd/broker/systemd_unit_test.go` asserts on
the shipped unit), and a `LoadConfig` unit test that chmods a temp config to 0644
and asserts an error (as `pkg/config/shipped_configs_test.go:82` asserts on
shipped configs).

### LINKED ISSUE #220: HIGH: installers insert pam_oidc.so at line 1 of the auth stack and wire console login through the broker, removing the documented recovery path
Found by the v0.5.0 security audit. Verified at `d8ca4b3`.

Both installers insert `pam_oidc.so` as the **first line** of the PAM auth stack,
and the shipped `configs/pam/login` has the same ordering problem. This makes the
broker a hard dependency of *every* authentication on the host, including the ones
that must keep working when the broker is down.

## Evidence

`scripts/install.sh` and `scripts/install-release.sh` both write the module line at
line 1 of the target PAM file, ahead of whatever was there. Nothing preserves the
existing stack's ordering constraints.

`configs/pam/login` has the same shape: `pam_oidc.so` precedes the local auth
modules.

The interaction that matters: `configs/pam/ssh` documents (`:108-115`) that the
recovery paths are pubkey auth and the console. If `login` also routes through
`pam_oidc.so` first, the console recovery path depends on the same broker.

## Failure scenario

The broker is down — because of the missing `RuntimeDirectory=` after a reboot
(filed separately), or a bad config, or the accept loop having exited on `EMFILE`.

SSH logins fail, which is the documented and accepted trade-off. But the operator's
**console login also goes through `pam_oidc.so` first**, so the documented recovery
path is gone too. The host is unreachable by every interactive route until someone
boots single-user or attaches a rescue image.

Separately, inserting at line 1 unconditionally can break stacks where an earlier
module is load-bearing — `pam_faillock`'s `preauth` entry must run before other
auth modules to count failures correctly, and on RHEL-family systems it is
ordinarily first.

## Fix

1. Never insert unconditionally at line 1. Insert **after** any `pam_faillock`
   preauth / `pam_env` style prologue, and use `auth sufficient` (as `configs/pam/ssh`
   already does) so a broker failure falls through rather than short-circuiting.
2. Leave `login` (console) out of the default wiring entirely, or wire it with an
   explicit local-auth fallback ahead of `pam_oidc.so`. The console must stay
   usable when the broker is not, because it is the documented recovery path.
3. Have the installer refuse to modify a stack it does not recognize, printing the
   line it would have added, rather than editing blind.

## Test

`test/e2e` can assert the resulting stack: after install, the console/`login` path
still authenticates with the broker stopped. That is the property that matters and
it is cheap to check in the existing container harness.

scttfrdmn:
Four defects in the install path. Each of them ends with a host nobody can log in to, or a secret every local user can read.

Closes #207, closes #208, closes #209, closes #220. **Closes #200** — the broker-side half landed in #239; this is the installer half it was left open for.

## #208 — the module was written where PAM never looks

Every release up to v0.4.2 put `pam_oidc.so` in `/lib/security`. That directory exists on neither Debian/Ubuntu (`/lib/<triplet>/security`) nor RHEL-family systems (`/lib64/security`), so the install aborted part-way through — *after* the binaries were in place — and if an operator created the directory to get past that, the module sat somewhere PAM never loads from.

That is not a no-op. With the stack this project ships, the unresolvable `sufficient` line fails and the login falls through to `auth requisite pam_deny.so`: every login refused. On a stack with a password fallback it means OIDC silently never runs while the operator believes it is enforced.

Both installers now ask dpkg/rpm where `pam_permit.so` is (a module libpam itself ships, so a directory holding it is a directory PAM loads from), fall back to probing the known layouts, and do it **before anything is written**. If they cannot find it they refuse to install. `make install` and `make install-dev` resolve it through the same implementation instead of hardcoding a path. `PAM_MODULE_DIR` overrides, with a warning that PAM only loads from its own compiled-in path.

## #220 — insertion at line 1, unconditionally

Both installers did `sed -i '2iauth sufficient pam_oidc.so'`. Ahead of a `pam_faillock.so preauth` entry that breaks failure counting on RHEL-family systems; ahead of `pam_nologin.so`, `pam_access.so` or `pam_securetty.so` it makes those gates unreachable, because a `sufficient` module that succeeds ends the auth phase before them.

They now walk the auth phase, allow a known prologue to stay in front, insert immediately before the first module that authenticates or terminates (`pam_unix.so` and friends, `pam_deny.so`, or an include/substack of a shared stack), and **refuse outright on a stack they do not recognise** — printing the exact line for the operator to place by hand. The added line is fenced in `# BEGIN oidc-pam` / `# END oidc-pam` so the uninstaller removes exactly what was added.

`scripts/install.sh` also stops editing PAM unless `--configure-pam` is passed. `install-release.sh` already required it; the two disagreeing is how one of them gets fixed and the other forgotten.

## #207 — the uninstaller locked the host out and reported success

`scripts/uninstall.sh` ran `sed -i '/pam_oidc.so/d' /etc/pam.d/sshd`. On the stack this repository itself ships (`configs/pam/ssh`: `auth sufficient pam_oidc.so` followed by `auth requisite pam_deny.so`) that leaves `pam_deny.so` as the entire auth phase — every SSH authentication refused, for every user, permanently — while printing `Uninstallation completed successfully!`. It edited sshd alone, so a host that had followed `configs/pam/README.md` was also left with `su` and `sudo` naming a module that had just been deleted, and no way to escalate and repair it.

It now, in order:

1. removes the OIDC lines from **every** `/etc/pam.d` file that references them, the fenced block by preference;
2. checks the resulting auth phase still has something that can authenticate, and restores the installer's pre-install backup if it does not (the old script wrote a backup and then never used it);
3. if there is no usable backup either, leaves the file **exactly as it was**, says so loudly, exits non-zero, and leaves `pam_oidc.so` installed so the stack keeps working. Removing the `.so` out from under a stack that still names it is the same lockout by another route.

## #209 — the config was world-readable and the check only warned

Both installers wrote `/etc/oidc-auth/broker.yaml` at `0644`: the AES-256 token encryption key and every client secret, readable by every local user on exactly the kind of multi-user host this product exists for. `LoadConfig` did check the mode — and logged one `WARNING` line to stderr and served anyway, having just validated the key for strength before publishing it. Its mask (`0137`) did not even catch group-read.

Now `0600 root:root` at install time, and **fatal** at load time (`OIDC_AUTH_DEV=true` overrides for development). A file mode is not something an operator can be warned about once at startup and expected to notice.

## #200, installer half

Neither installer creates an `oidc-auth` account any more, and nothing is chowned to one. The broker runs as `User=root` — it has to, since on a successful login it writes into an arbitrary account's `~/.ssh` and hands the file over — so an unprivileged account owning the socket directory is a privilege boundary pointing the wrong way: `unlink(2)` + `bind(2)` installs an impostor at the same path that every PAM login then talks to, while the real broker serves on its unlinked inode and stays green in `systemctl status`.

`$LOG_DIR` and `broker.log` were chowned the same way, which is a root process appending to a file an unprivileged account can replace with a symlink. Both are now root-owned. A leftover account from an earlier install produces a warning, not a silent `userdel`.

## Adjacent, and deliberately in the same lines

- **`uninstall.sh` sweeps what the broker left behind** — the `authorized_keys` entries it issued (recognised exactly as `pkg/ssh/entry.go` recognises them) and `/var/lib/oidc-pam`, which holds the private keys. Neither was touched before, so every key the broker had ever issued survived an uninstall. The sweep refuses a `~/.ssh` or `authorized_keys` that is a symlink, or that belongs to neither the account nor root: root following a user-controlled path is #204/#225 by another route. The residual check-then-use window is called out in a comment — a shell has no `openat(2)`.
- **`test/scripts/test-pam-wiring.sh`** exercises both scripts' PAM logic against a throwaway `/etc/pam.d` — no root, no PAM headers, nothing installed — and CI now runs it plus `shellcheck --severity=warning` over `scripts/`. 32 assertions. Neutering `pam_auth_phase_authenticates` makes the #207 case fail, so the guard is not vacuous. Two pre-existing shellcheck warnings in unrelated scripts are fixed so the gate can cover all of `scripts/`.
- **`configs/pam/login`** had `pam_nologin.so` and `pam_group.so` *below* a `sufficient` line, so for any user OIDC admitted, `/etc/nologin` was ignored and `pam_group` never ran at all. Reordered, with plain control flags and no numeric jumps — `[success=N default=ignore]` encodes a line count that silently changes meaning when a line is added, which is the #122/#172 fragility.
- **Docs**: `DEPLOYMENT.md`, `REQUIREMENTS.md` and `configs/pam/README.md` all documented the `/lib/security` path and the `oidc-auth` ownership this change removes, so a host built by following the guide carried both defects.

## Not fixed here

`scripts/build-releases.sh` and `scripts/build-simple.sh` each emit their own installer into a tarball, carrying all four of these defects. Nothing references either script — not the Makefile, not CI, not the docs; `release.yml` packages `scripts/install-release.sh`. Rewriting two dead scripts is worse than filing it, so it is filed separately.

## Verification

- `go build ./... && go vet ./... && go test ./...` clean; `gofmt -s -l .` empty.
- `test/scripts/test-pam-wiring.sh`: 32/32.
- `shellcheck --severity=warning scripts/*.sh test/scripts/*.sh` clean.
- Non-vacuity: with `pam_auth_phase_authenticates` forced to succeed, the #207 case fails on 3 of its 6 assertions; the file was restored byte-exact afterwards (`diff -q`).
- The `make install` recipe expansion was checked with `make -n`, and the resolver verified to abort with the `PAM_MODULE_DIR` advice on a host with no libpam (this macOS one) and to honour the override.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -271,3 +271,16 @@ jobs:
             exit 1
           fi
           echo "release.yml's build job is gated on the CI workflow."
+
+      # The installer and uninstaller decide, in shell, whether a host stays
+      # loggable-in: #207 was an uninstaller that reduced /etc/pam.d/sshd to
+      # `auth requisite pam_deny.so` and reported success. This exercises that
+      # logic against a throwaway /etc/pam.d — no root, no PAM headers, nothing
+      # installed — so the guard is not left to a reviewer's reading.
+      - name: Check the installer PAM wiring
+        run: ./test/scripts/test-pam-wiring.sh
+
+      - name: Shellcheck the install scripts
+        run: |
+          sudo apt-get update -qq && sudo apt-get install -y -qq shellcheck
+          shellcheck --severity=warning scripts/*.sh test/scripts/*.sh
--- a/DEPLOYMENT.md
+++ b/DEPLOYMENT.md
@@ -242,10 +242,18 @@ systemctl status oidc-auth-broker
 ```
 
 The bundled `install.sh` places `oidc-auth-broker`, `oidc-pam-helper`, and
-`oidc-admin` in `/usr/local/bin`, `pam_oidc.so` in `/lib/security`, an example
-config in `/etc/oidc-auth/broker.yaml`, and the systemd unit in
-`/etc/systemd/system`. It does **not** start the broker or modify PAM unless
-`--configure-pam` is supplied, so it cannot lock you out on its own.
+`oidc-admin` in `/usr/local/bin`, `pam_oidc.so` in whichever directory *this
+host's* libpam loads modules from, an example config in
+`/etc/oidc-auth/broker.yaml` (mode `0600`, `root:root` — the broker refuses to
+start otherwise), and the systemd unit in `/etc/systemd/system`. It does **not**
+start the broker or modify PAM unless `--configure-pam` is supplied, so it cannot
+lock you out on its own.
+
+The module directory is detected, not assumed: it is `/lib/<triplet>/security` on
+Debian/Ubuntu and `/lib64/security` on RHEL-family systems, and there is no
+`/lib/security` on either, which is where every release up to v0.4.2 wrote it
+(#208). Set `PAM_MODULE_DIR` if detection gets it wrong; the installer refuses
+rather than guessing, because a module outside that path never loads at all.
 
 ### Method 3: Source Compilation
 
@@ -260,6 +268,11 @@ cd oidc-pam
 make build
 sudo make install
 ```
+
+`make install` detects libpam's module directory the same way the installers do
+and refuses if it cannot find it; set `PAM_MODULE_DIR` to override. It installs no
+PAM configuration at all — use `sudo ./scripts/install.sh --configure-pam` if you
+want `/etc/pam.d/sshd` wired for you, and read "PAM Integration" below first.
 
 ## Configuration
 
@@ -542,19 +555,45 @@ reason is recorded as an `ssh_key_provisioning_failed` audit event.
 
 ### 1. File Permissions
 
+Everything the broker reads or writes is root-owned, because the broker runs as
+root. There is no `oidc-auth` service account — nothing runs as one, and earlier
+versions of this guide and of `scripts/install.sh` created one and gave it the
+socket and log directories. That was a privilege boundary pointing the wrong way:
+write access to the directory holding the socket is enough to unlink the socket and
+bind an impostor at the same path, which every PAM login then talks to (#200), and
+a log file an unprivileged account owns is one it can replace with a symlink for a
+root process to append through. If you installed an earlier version, re-run the
+installer or apply the commands below; the broker now refuses to serve from a
+directory anyone else can write to.
+
 ```bash
-# Set secure permissions
-sudo chown -R oidc-auth:oidc-auth /etc/oidc-auth
+# Configuration: read by the root broker, and holds the OIDC client secret
+sudo chown -R root:root /etc/oidc-auth
 sudo chmod 750 /etc/oidc-auth
 sudo chmod 640 /etc/oidc-auth/broker.yaml
 
-# Log directory permissions
-sudo chown -R oidc-auth:oidc-auth /var/log/oidc-auth
+# Log directory
+sudo chown -R root:root /var/log/oidc-auth
 sudo chmod 750 /var/log/oidc-auth
 
-# PAM module permissions
-sudo chown root:root /lib/security/pam_oidc.so
-sudo chmod 644 /lib/security/pam_oidc.so
+# Socket directory. systemd's RuntimeDirectory= reasserts this on every start;
+# these are the values a hand-built host needs before the first one.
+sudo chown root:root /run/oidc-auth
+sudo chmod 750 /run/oidc-auth
+
+# Broker state: the SSH keys it issues and the locks serializing its
+# authorized_keys writes. Created by StateDirectory= in the unit.
+sudo chown -R root:root /var/lib/oidc-pam
+sudo chmod 700 /var/lib/oidc-pam
+
+# PAM module. Install it in the directory libpam itself loads from, which differs
+# by distribution -- /lib/<triplet>/security on Debian/Ubuntu, /lib64/security on
+# RHEL-family. `scripts/install.sh` detects it; if you place the module by hand,
+# find it with the directory holding pam_permit.so. A module outside that path
+# never loads at all (#208).
+PAM_DIR="$(dirname "$(find /lib /usr/lib -name pam_permit.so -print -quit)")"
+sudo chown root:root "$PAM_DIR/pam_oidc.so"
+sudo chmod 644 "$PAM_DIR/pam_oidc.so"
 ```
 
 ### 2. Network Security
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,28 @@ endif
 # CI Lint job, the verification container and a local `make lint` cannot disagree.
 GOLANGCI_LINT_VERSION := $(shell tr -d '[:space:]' < .golangci-version)
 
+# Resolve the directory this host's libpam loads modules from, into $$pamdir.
+#
+# `make install` used to copy the module to /lib/security, which exists on neither
+# Debian/Ubuntu (/lib/<triplet>/security) nor RHEL-family (/lib64/security)
+# systems (#208) -- so the copy failed, or, once an operator created the directory
+# to get past that, put the module somewhere PAM never looks and every login on
+# the shipped stack was refused by pam_deny.so instead.
+#
+# scripts/install.sh's detect_pam_dir is the one implementation of the probe;
+# sourcing that script defines its functions and runs nothing. Override with
+# PAM_MODULE_DIR, exactly as the installers do.
+define resolve_pam_dir
+pamdir="$${PAM_MODULE_DIR:-$$(bash -c 'source ./scripts/install.sh; detect_pam_dir' || true)}"; \
+if [ -z "$$pamdir" ]; then \
+	echo "Could not find the directory libpam loads modules from. Install your" >&2; \
+	echo "distribution's PAM modules (libpam-modules on Debian/Ubuntu, pam on" >&2; \
+	echo "RHEL-family), or set PAM_MODULE_DIR to the directory holding pam_unix.so." >&2; \
+	exit 1; \
+fi; \
+echo "  PAM module directory: $$pamdir"
+endef
+
 # Default target
 all: build
 
@@ -119,10 +141,11 @@ verify-linux:
 ## Install binaries to system locations
 install: build
 	@echo "Installing binaries..."
-	sudo cp $(BINARY_DIR)/$(PAM_MODULE) /lib/security/
-	sudo cp $(BINARY_DIR)/$(BROKER_BINARY) /usr/local/bin/
-	sudo cp $(BINARY_DIR)/$(HELPER_BINARY) /usr/local/bin/
-	sudo cp $(BINARY_DIR)/$(ADMIN_BINARY) /usr/local/bin/
+	@set -e; $(resolve_pam_dir); \
+	sudo install -m 0644 $(BINARY_DIR)/$(PAM_MODULE) "$$pamdir/$(PAM_MODULE)"
+	sudo install -m 0755 $(BINARY_DIR)/$(BROKER_BINARY) /usr/local/bin/$(BROKER_BINARY)
+	sudo install -m 0755 $(BINARY_DIR)/$(HELPER_BINARY) /usr/local/bin/$(HELPER_BINARY)
+	sudo install -m 0755 $(BINARY_DIR)/$(ADMIN_BINARY) /usr/local/bin/$(ADMIN_BINARY)
 	sudo cp configs/systemd/oidc-auth-broker.service /etc/systemd/system/
 	sudo systemctl daemon-reload
 	sudo systemctl enable oidc-auth-broker
@@ -130,12 +153,15 @@ install: build
 ## Install development version
 install-dev: build
 	@echo "Installing development version..."
-	sudo cp $(BINARY_DIR)/$(PAM_MODULE) /lib/security/
-	sudo cp $(BINARY_DIR)/$(BROKER_BINARY) /usr/local/bin/
-	sudo cp $(BINARY_DIR)/$(HELPER_BINARY) /usr/local/bin/
-	sudo cp $(BINARY_DIR)/$(ADMIN_BINARY) /usr/local/bin/
-	sudo mkdir -p /etc/oidc-auth
-	sudo cp configs/examples/broker.yaml /etc/oidc-auth/
+	@set -e; $(resolve_pam_dir); \
+	sudo install -m 0644 $(BINARY_DIR)/$(PAM_MODULE) "$$pamdir/$(PAM_MODULE)"
+	sudo install -m 0755 $(BINARY_DIR)/$(BROKER_BINARY) /usr/local/bin/$(BROKER_BINARY)
+	sudo install -m 0755 $(BINARY_DIR)/$(HELPER_BINARY) /usr/local/bin/$(HELPER_BINARY)
+	sudo install -m 0755 $(BINARY_DIR)/$(ADMIN_BINARY) /usr/local/bin/$(ADMIN_BINARY)
+	sudo install -d -m 0750 -o root -g root /etc/oidc-auth
+	# 0600 root:root: this file holds the token encryption key and every client
+	# secret, and the broker refuses to start if anyone else can read it (#209).
+	sudo install -m 0600 -o root -g root configs/examples/broker.yaml /etc/oidc-auth/broker.yaml
 	sudo cp configs/systemd/oidc-auth-broker.service /etc/systemd/system/
 	sudo systemctl daemon-reload
 
--- a/REQUIREMENTS.md
+++ b/REQUIREMENTS.md
@@ -77,12 +77,16 @@ All Go dependencies are managed through `go.mod` and will be automatically downl
 - `/etc/oidc-auth/` directory for configuration
 - `/var/run/oidc-auth/` directory for runtime files (sockets, PIDs)
 - `/var/log/oidc-auth/` directory for logs (optional)
-- `/lib/security/` directory for PAM module
+- `/var/lib/oidc-pam/` directory for broker state (issued SSH keys, write locks)
+- the directory this host's libpam loads modules from, for `pam_oidc.so` —
+  `/lib/<triplet>/security` on Debian/Ubuntu, `/lib64/security` on RHEL-family.
+  There is no `/lib/security` on either (#208); `scripts/install.sh` detects the
+  right one and refuses to install if it cannot
 - `/usr/local/bin/` directory for binaries
 
 ### User Permissions
 - Root privileges required for:
-  - Installing PAM module to `/lib/security/`
+  - Installing PAM module into libpam's module directory
   - Configuring PAM in `/etc/pam.d/`
   - Installing systemd service files
   - Creating system directories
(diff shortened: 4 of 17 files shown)


scttfrdmn:
Pushed a fix for the end-to-end failure on the last run, which was the change working as intended: `test/e2e/Dockerfile.broker:43` used a bare `COPY`, whose default mode is 0644, so the broker refused to start on its own harness config —

```
{"level":"fatal","error":"config file /etc/oidc-auth/broker.yaml has permissions 0644: it holds the token encryption key and every client secret, so it must not be readable by anyone but root ..."}
```

Now `COPY --chmod=600`. Deliberately **not** `OIDC_AUTH_DEV=true`: this container starts the broker the way a host starts it, so it is the one place the fatal check has to be exercised rather than opted out of. That was the only `broker.yaml` install under `test/` (`grep -rn broker.yaml test/`); `test/e2e/Dockerfile.client` already asks dpkg for the PAM module directory rather than hardcoding one.

Issue #140, opened 2026-08-13
pam_oidc.so is built with no PAM entry points: the module cannot load, and every C-side fix is absent from the artifact


scttfrdmn:
`make build-pam` produces a `pam_oidc.so` that contains **none** of the six `pam_sm_*` entry points. PAM cannot load a module without them, so the module has never been functional in any released artifact. The build succeeds silently.

## Evidence

```
$ CGO_ENABLED=1 go build -buildmode=c-shared -o pam_oidc.so ./cmd/pam-module
$ echo $?
0
$ nm -D --defined-only pam_oidc.so | grep -c " T pam_sm_"
0
$ readelf -d pam_oidc.so | grep NEEDED
 0x0000000000000001 (NEEDED)  Shared library: [libc.so.6]
```

`libpam` is absent from `NEEDED` despite `#cgo LDFLAGS: -lpam` in `cmd/pam-module/main.go` — the linker dropped it as unused, which independently confirms no PAM code is in the object. The six entry points do exist in the source:

```
$ grep -o 'pam_sm_[a-z_]*' pkg/pam/cgo_bridge.c | sort -u
pam_sm_acct_mgmt  pam_sm_authenticate  pam_sm_chauthtok
pam_sm_close_session  pam_sm_open_session  pam_sm_setcred
```

## Cause

`cgo_bridge.c` lives in `pkg/pam/`. **cgo only compiles the C sources that sit in the directory of the package being built.** `cmd/pam-module/main.go` includes `cgo_bridge.h` via `-I${SRCDIR}/../../pkg/pam`, which supplies *declarations only*, and `cmd/pam-module` does not import `pkg/pam`:

```
$ grep -rn 'oidc-pam/pkg/pam' cmd/pam-module/
(no matches)
```

So the entry points are compiled into the `pkg/pam` archive — where nothing links them into a shared object — and `cmd/pam-module` builds a `.so` from a `main` package whose only content is an empty `main()`.

Header-only linkage is why this is silent: the compiler sees valid declarations, the linker is asked for symbols nobody references, and `-Wl,--as-needed` quietly discards `-lpam`.

## Impact

- `make install` / `make install-dev` copy this `.so` to `/lib/security/`, and the release tarballs ship it. A PAM stack referencing it gets a module-load failure, not authentication.
[...]

Issue #241, opened 2026-08-14
scripts/build-releases.sh and build-simple.sh each emit a third installer, carrying #207/#208/#209/#200


scttfrdmn:
`scripts/build-releases.sh:223` and `scripts/build-simple.sh:189` each write their own `install.sh` into the tarball they build, by heredoc. Both of those emitted installers do:

```bash
cp pam_oidc.so /lib/security/
chmod 644 /lib/security/pam_oidc.so
```

and go on to create the directories at the old modes. So they carry, in a copy nobody looks at:

- **#208** — `/lib/security` exists on neither Debian/Ubuntu (`/lib/<triplet>/security`) nor RHEL-family (`/lib64/security`) systems. The `cp` fails, or — if the operator creates the directory to get past it — puts the module somewhere PAM never loads from, which on the shipped stack means every login is refused by `pam_deny.so`.
- **#209** — the config is written world-readable.
- **#200** — the socket and log directories are handed to an `oidc-auth` account.
- **#207/#220** — no fenced insertion, no refusal on an unrecognised stack, and no matching uninstall path.

## Why this was not fixed alongside the rest

Nothing references either script. Not the `Makefile`, not `.github/workflows/`, not the docs, not `scripts/release.sh` — `release.yml:142` packages `scripts/install-release.sh` as the tarball's `install.sh`, which is the one that got fixed. `grep -rn 'build-releases.sh\|build-simple.sh'` outside the two files themselves returns nothing.

Rewriting two unreferenced scripts to re-emit a fourth and fifth copy of installer logic is worse than the duplication it would replace, so it was left out of that PR deliberately rather than silently.

## What to do

Pick one, they are all fine:

1. **Delete both.** They predate `scripts/release.sh` and `install-release.sh` and nothing invokes them. This is the recommendation.
2. Have them `cp scripts/install-release.sh "$PKG/install.sh"` instead of emitting a heredoc, so there is one installer in the repository and no copy to drift.
3. Keep them and fix the emitted installer, accepting a third copy of the PAM-directory probe.

Whichever: a `grep -rn '/lib/security' scripts/` that comes back empty is the check that this cannot recur, and it belongs in the CI `Validate` job next to `test/scripts/test-pam-wiring.sh`.

## Severity

Low — no shipped artifact is affected, because the release workflow does not use these scripts. It is a landmine for whoever runs one by hand and hands the tarball to someone else.

scttfrdmn:
[...]

PR #252, opened 2026-08-14
build(pam): build the module as hardened C, from the one script that verifies it (#198, #222)

### LINKED ISSUE #198: The PAM module is compiled without hardening flags, and carries the Go runtime's signal handlers into sshd
Found reviewing oauth2-pam's equivalent; both halves apply here. Filed from the oauth2-pam side, no changes made to this repository. The second half may be unfixable for you — I'd rather you decide that knowingly.

**1. No hardening flags.** `cmd/pam-module/bridge_linux.go:6` is:
```go
#cgo CFLAGS: -I${SRCDIR} -I/usr/include/security -Wall -Wextra
```
`-Wall -Wextra` is more than oauth2-pam had, but `git grep` finds no `-Werror`, no `-fstack-protector-strong`, no `-D_FORTIFY_SOURCE`, and no `-Wl,-z,relro -Wl,-z,now` anywhere in the repository. The artifact is a shared object loaded into `sshd` as root, holding a 16 KB response buffer on the stack (`:656`). These are cheap to add and, since the code is warning-clean, cost nothing today — which is the argument for doing it before something needs fixing to enable them.

**2. `-buildmode=c-shared` gives the Go runtime your signal dispositions.** `Makefile:63` (and `:207`, `:211`, and `release.yml:107`) build with `-buildmode=c-shared`, so every `sshd` pre-auth child that loads the module starts the Go runtime, which unconditionally installs its own handlers for **SIGSEGV, SIGBUS, SIGFPE, SIGPIPE and SIGURG**, plus `DF_1_NODELETE` so `dlclose` cannot unload it.

Unlike oauth2-pam — whose module has *no* Go code, so it can drop `c-shared` for a plain `cc -shared -fPIC` build (scttfrdmn/oauth2-pam#65) — this repository has real Go logic in the module (`args_linux.go`, `auth_linux.go`, `pkg/pam`), so that exit is not available to you without a rewrite. The options I can see are: accept it and document it next to the module's other host-impact notes; or move the remaining logic to C and take the same exit. I'm not arguing for either — but it's a property of the shipped artifact that isn't written down anywhere, and it interacts with anything that reasons about the module's effect on the host process.

A related benefit if you ever do take the C route: a plain C build is analyzable by a CodeQL `c-cpp` database with an ordinary build command, which is the gap oauth2-pam tracks as its #38.

### LINKED ISSUE #222: MEDIUM: three of four pam_oidc.so producers skip verify-pam-module.sh, and build-simple.sh emits a mismatched architecture
Found by the v0.5.0 security audit. Verified at `d8ca4b3`.

`scripts/verify-pam-module.sh` exists because a `-buildmode=c-shared` build with
**no PAM entry points in it exits 0** (#140) — the artifact has to be inspected or
a useless module ships. `release.yml:115-116` runs it.

**Three other producers of `pam_oidc.so` do not**, and one of them builds for the
wrong architecture.

## Evidence

`release.yml:112-116` states the rule in its own comment:

```
# A c-shared build with no PAM entry points in it exits 0 (#140), so the
# artifact has to be inspected before it can be released. This step is what
# stops another release shipping a module PAM cannot use.
```

The three paths that build the module and skip that check:

1. **`scripts/build-simple.sh`** — builds the `.so` with no `verify-pam-module.sh`
   call, **and** produces a binary whose architecture does not match the target it
   claims: it does not set `GOARCH` consistently with the output name, so on an
   arm64 host it emits an arm64 `.so` under a name consumers read as the amd64
   artifact.
2. **The `Makefile`** target that builds the module — no verify step.
3. **The e2e harness** build path — no verify step, which is the one that matters
   most: the e2e suite is meant to be the thing that catches a broken module, and
   it will happily test a module with no entry points, since the failure mode of
   #140 is a module PAM cannot load at all.

## Failure scenario

A contributor changes a cgo export signature or a build tag such that the PAM entry
points stop being emitted. `go build` exits 0. `make` exits 0. The e2e suite builds
the module and — depending on how the stack is wired — reports failures that read as
test flakiness rather than "the module has no entry points".

Nothing surfaces the real cause until someone reads `nm` output by hand, which is
precisely the debugging session `verify-pam-module.sh` was written to prevent.

The `build-simple.sh` arch mismatch is worse in a different way: it produces a file
that installs cleanly and then fails at `dlopen` time on the target host, inside
sshd's auth path.

## Fix

1. Call `scripts/verify-pam-module.sh` from **every** path that produces
   `pam_oidc.so`: `build-simple.sh`, the `Makefile` target, and the e2e build.
2. Fix `build-simple.sh` to set `GOARCH` explicitly and name its output for the
   architecture it actually built, or refuse to cross-build at all.
3. The general fix is to have exactly one place that builds the module and have
   everything else call it, so the verify step cannot be forgotten again. Four
   producers is the root cause; the missing check is the symptom.

## Test

The verify script *is* the test. What is missing is calling it. A CI check that
greps for `c-shared` in the repo and asserts each occurrence is accompanied by a
`verify-pam-module.sh` call would pin it.

scttfrdmn:
`pam_oidc.so` is dlopened as root into every sshd pre-auth child. It was compiled
without hardening flags, and it was built as a Go `c-shared` library, which put
the Go runtime into that process. Both are fixed here, and the reason they land in
one PR is that the fix for #222 — one build path instead of several — is what makes
either change apply everywhere the module is produced.

## The producer inventory, established here

`git grep -n buildmode=c-shared` on `main`, over `Makefile`, `scripts/`,
`.github/workflows/`, `test/` and `docs/`, finds **eight build sites in six
files**, not four:

| Build site (`main`) | Ran `verify-pam-module.sh`? |
| --- | --- |
| `Makefile:85` — `build-pam` | yes (`Makefile:87`) |
| `Makefile:138` — `verify-linux` container sweep | yes (same `sh -c` chain) |
| `Makefile:233` — `release`, amd64 | **no** |
| `Makefile:237` — `release`, arm64 | **no** |
| `.github/workflows/release.yml:172` | yes (next step) |
| `test/e2e/Dockerfile.client:22` | yes (`:25`) |
| `scripts/build-simple.sh:84` | **no** |
| `scripts/build-releases.sh:110` | **no** |

(`docs/design/oidc_pam_project.md:296` and `scripts/verify-pam-module.sh:11` also
contain the string; both are prose, and the design doc's is a stale
`./pkg/pam` snippet — see "Found, not fixed".)

**So #222 is wrong on two of the three paths it names.** It lists `build-simple.sh`,
"the `Makefile` target that builds the module", and "the e2e harness build path" as
the three skippers, and calls the e2e one "the one that matters most". In fact
`make build-pam` and the e2e image were both verifying already; the paths that
weren't are the `Makefile`'s `release` target (both architectures) and
`scripts/build-releases.sh`, neither of which the issue mentions. Its
count is low and its set is wrong, but its point 3 — "the general fix is to have
exactly one place that builds the module" — is right, and is what this PR does.

Its architecture claim is right about the consequence and wrong about the cause.
`build-simple.sh` did not misname its build: it wrote `pam_oidc-native.so`, which
claims no architecture. The bug is in the *packaging* loop below it, which copied
that single file into **every** Linux tarball, so on an arm64 host the
`linux-amd64` package shipped an arm64 module — installs cleanly, fails at
`dlopen` inside sshd's auth path. The PAM helper was packaged the same way.

## What #198's second half assumed, and what is actually true

The issue says the `c-shared` exit "is not available to you without a rewrite",
because "this repository has real Go logic in the module (`args_linux.go`,
`auth_linux.go`, `pkg/pam`)". Verified otherwise:

- `grep -rn '//export' cmd/ pkg/` → nothing. There is no C→Go entry point.
- All six `pam_sm_*` functions are defined in `cmd/pam-module/cgo_bridge_linux.c`.
  `nm -D --defined-only` on the shipped `.so` finds them there.
- `cgo_bridge_linux.c` and `cgo_bridge.h` reference no Go symbol, no `_cgo_*`
  thunk, and no `GoString`/`GoBytes`.
- The package's Go files are cgo *test* wrappers — cgo is not allowed in
  `_test.go`, so they live in normal files and `go test` drives the C through
  them. They are not part of the module.

No rewrite was needed, so the module is now built by handing that one C file to
the C compiler.

## Claims verified, and how

Everything below was measured inside Linux containers, reusing the existing paths
(`test/docker/Dockerfile.verify` and `test/e2e/`) rather than a new image. **This
host is macOS and cannot build the module at all** — the C bridge needs
`<security/pam_ext.h>` and `<json-c/json.h>`.

Against the module built by `scripts/build-pam-module.sh`:

- `readelf -lW` → `GNU_RELRO` present; `readelf -dW` → `BIND_NOW` + `FLAGS_1: NOW`
  (full RELRO). The old `c-shared` build had `GNU_RELRO` without `BIND_NOW`, i.e.
  partial RELRO with a writable PLT.
- `GNU_STACK` is `RW`, not `RWE`.
- `nm -D --undefined-only` → `__stack_chk_fail` present (was absent), and
  `__memcpy_chk`/`__snprintf_chk`/`__vsnprintf_chk` (fortify active, so `-O2` is
  reaching it).
- No `TEXTREL`.
- All six `pam_sm_*` exported; `libpam.so.0` and `libjson-c.so` in `NEEDED`.
- No `runtime.*`/`go:` symbols, **no `pthread_create`, no `sigaction`**, no
  `DF_1_NODELETE`. Size 72 KB against 2.4 MB.
- Warning-clean under `-Wall -Wextra`.

For #198's signal claim on the *old* build, I read the Go runtime source rather
than infer: `initsig` runs at library init, `sigInstallGoHandler` in library mode
installs handlers for `_SigPanic` signals plus SIGPIPE and `sigPreempt`, and
`sigtab_linux_generic.go` marks SIGSEGV/SIGBUS/SIGFPE as `_SigPanic` with
`sigPreempt = SIGURG` — the five the issue names, exactly. `readelf -dW` on a
`c-shared` artifact confirmed `FLAGS_1: NODELETE`. The runtime also adds
`SA_ONSTACK` to sshd's pre-existing handlers via `setsigstack` and forwards
through `sigfwdgo`, so the effect is not a clean replacement.

Behavioural verification, which is the one that mattered for switching build
modes: **`./test/e2e/run-tests.sh` → 17/17 cases passed** against the C-built,
hardened module. That harness runs a real `sshd` with a real PAM stack, the real
broker and a fake IdP, and it covers the failure cases (`broker_down`, key
supersession, orphan sweeps, per-account rate limiting), so the module is loaded
by libpam and exercised end to end, not merely inspected.

Also verified in a container: `check-cgo-quarantine.sh`,
`check-pam-module-producers.sh`, `go vet ./cmd/pam-module`,
`go test ./cmd/pam-module/...`, and the arch assertion in both directions (a
matching `expected-arch` is accepted, a mismatched one exits non-zero). I broke
each of the three failure branches of `check-pam-module-producers.sh` on purpose
and confirmed each fires — including one real bug that testing found: its "the
builder still verifies" assertion originally passed on a builder with the verify
call deleted, because `grep -q` matched the builder's own explanatory comments.

Locally on macOS: `go build ./...`, `go vet ./...`, `gofmt`,
`check-cgo-quarantine.sh`, `check-pam-module-producers.sh`,
`shellcheck --severity=warning scripts/*.sh test/scripts/*.sh` (the severity CI
uses), `bash -n` on every touched script, and `yaml.safe_load` on both edited
workflows. No `uses:` line is touched, so SHA pinning is unaffected.

## Only verifiable in CI

- The release build on an **arm64** runner. The arch assertion and the aarch64
  branch of the machine-name mapping have not run on real arm64 hardware here;
  the `Build and verify the PAM module` step in `release.yml`'s matrix is what
  exercises it. It fails loudly rather than silently if the mapping is wrong.
- The **`PAM (cgo)`** job compiling the C through cgo with the new `#cgo`
  hardening flags on Ubuntu 24.04, where `_FORTIFY_SOURCE` already defaults to 3;
  the floor-not-assignment form is what should keep that at 3 without a
  redefinition warning. I tested this on the Debian-based verify image, not on
  24.04.
- `make release` and `scripts/build-releases.sh` end to end. Neither can complete
  on any single host (see below), so what I verified is the module line, not the
  whole target.

## Consequences worth knowing

- **The `#cgo` flags and the script's flags are now two lists that must agree.**
  A C build has no way to share them. Mitigation is adjacency plus comments in
  both, a note in CONTRIBUTING, and the fact that the e2e suite exercises the
  script's artifact — so a drift that breaks the module is caught, though a drift
  that only weakens the tests' copy is not.
- `bin/pam_oidc.so` no longer carries `-X main.version=...`. It never did anything:
  the package declares no such variables, so the linker discarded it.
- `make release` still cannot produce a module for the non-host architecture — it
  now says so instead of failing inside the compiler. Releases come from
  `release.yml`'s per-architecture runners.

## Found, not fixed

- `scripts/build-simple.sh` and `scripts/build-releases.sh` generate installer
  scripts inside their tarballs that still copy `pam_oidc.so` to `/lib/security`,
  the hardcoded path #208 removed from the real installers. That directory exists
  on neither Debian/Ubuntu nor RHEL-family systems, so a tarball's own installer
  reintroduces exactly the failure #208 fixed.
- Neither `make release` nor `scripts/build-releases.sh` can complete on any one
  host: the pure-Go cross-builds are fine, but the cgo `pam-helper` and the module
  are not. They read as full release paths and are not.
- `make docker-build` references a root `Dockerfile` that does not exist in the
  repository.
- `docs/design/oidc_pam_project.md:296` shows
  `go build -buildmode=c-shared -o … ./pkg/pam`, which is both the pre-#140 layout
  and now the wrong build mode. That file is under the unmaintained design
  directory, so I left it alone.

Closes #198
Closes #222

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,6 +233,14 @@ jobs:
       - name: Check the cgo quarantine
         run: ./scripts/check-cgo-quarantine.sh
 
+      # #222: four places compiled pam_oidc.so and three of them did not inspect
+      # the result, which a build with no PAM entry points in it does not report
+      # (#140). scripts/build-pam-module.sh cannot finish without inspecting it, so
+      # what has to stay true is that nothing else compiles the module — a second
+      # producer is where the check gets forgotten again.
+      - name: Check one script is the only producer of the PAM module
+        run: ./scripts/check-pam-module-producers.sh
+
       # Every `uses:` in this directory must name a 40-hex commit SHA. A version
       # tag is mutable: whoever controls the action's repository can retarget it,
       # and the two references that were left unpinned sat between "build the PAM
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -163,22 +163,15 @@ jobs:
             -o bin/oidc-admin-linux-${{ matrix.arch }} \
             ./cmd/oidc-admin
 
-      - name: Build PAM module (CGO)
-        env:
-          GOOS: linux
-          GOARCH: ${{ matrix.arch }}
-          CGO_ENABLED: 1
-        run: |
-          go build -buildmode=c-shared -trimpath -ldflags="${{ steps.vars.outputs.ldflags }}" \
-            -o bin/pam_oidc.so-linux-${{ matrix.arch }} \
-            ./cmd/pam-module
-          rm -f bin/pam_oidc.so-linux-${{ matrix.arch }}.h
-
-      # A c-shared build with no PAM entry points in it exits 0 (#140), so the
-      # artifact has to be inspected before it can be released. This step is what
-      # stops another release shipping a module PAM cannot use.
-      - name: Verify the PAM module is loadable
-        run: ./scripts/verify-pam-module.sh bin/pam_oidc.so-linux-${{ matrix.arch }}
+      # The module is C, built by the one script every producer uses, which
+      # applies the hardening flags and then inspects what it produced: a build
+      # with no PAM entry points in it exits 0 (#140), so nothing but reading the
+      # artifact can tell a release from a module PAM cannot use. The architecture
+      # argument makes the file name a claim the script checks — this job runs on a
+      # runner of the matrix architecture, and a change that made it cross-build
+      # would otherwise publish a module that fails at dlopen (#222).
+      - name: Build and verify the PAM module
+        run: ./scripts/build-pam-module.sh bin/pam_oidc.so-linux-${{ matrix.arch }} ${{ matrix.arch }}
 
       - name: Build PAM helper (CGO)
         env:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ## [Unreleased]
 
 ### Security
+- **`pam_oidc.so` no longer carries the Go runtime into `sshd`, and is compiled
+  with hardening flags (#198).** The module is loaded into every authenticating
+  `sshd` pre-auth child, as root, and holds a 16 KB broker response on the stack.
+  It was compiled with `-Wall -Wextra` and nothing else: no stack protector, no
+  `_FORTIFY_SOURCE`, and partial RELRO with a writable PLT.
+
+  It was also built as a Go `c-shared` library, which starts the Go runtime in
+  every process that `dlopen`s it. The runtime replaces the process's handlers for
+  `SIGSEGV`, `SIGBUS`, `SIGFPE`, `SIGPIPE` and `SIGURG` with its own, adds
+  `SA_ONSTACK` to the handlers it leaves alone, runs its own threads, and marks the
+  library `DF_1_NODELETE` so `dlclose` cannot unload it. It did all of that in
+  `sshd` in service of no Go code: all six `pam_sm_*` entry points are C, they call
+  no Go, and the package exports no Go function to C, so the Go files in
+  `cmd/pam-module` are cgo wrappers that exist to let `go test` drive the C.
+
+  The module is now compiled from that one C file by the C compiler, with
+  `-fstack-protector-strong`, a `_FORTIFY_SOURCE` floor of 2 set in the source so a
+  distribution asking for a higher level keeps it, and `-Wl,-z,relro -Wl,-z,now
+  -Wl,-z,noexecstack`. The result is 70 KB rather than 2.4 MB, has full RELRO and a
+  guarded stack, starts no threads, installs no signal handlers, and can be
+  unloaded. Every SSH login in the end-to-end harness authenticates against it.
+
+  The same flags are in `bridge_linux.go`'s `#cgo` directives, so the C the tests
+  exercise is compiled the way the shipped module is. `-Werror` was deliberately
+  not added: the code is warning-clean today, but making a future compiler's new
+  diagnostic a build failure for anyone building from source is a different bargain
+  from the flags above, and CI compiles the module on every push anyway.
 - **Release artifacts are now signed and carry build provenance (#180).** Until
   now the only integrity material published was a `.sha256` file uploaded to the
   same release page as the tarball it describes, which detects a corrupted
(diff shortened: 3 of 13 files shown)


scttfrdmn:
Rebased onto `main` after #248 landed. #248 deletes `scripts/build-simple.sh` and `scripts/build-releases.sh`; this PR modified both, so the rebase conflicted on exactly those two files and I resolved it by keeping the deletion — the edits were about making those two scripts call `build-pam-module.sh`, and a deleted script needs no producer discipline.

Three consequences, handled rather than left:

- `scripts/check-pam-module-producers.sh` needed no change. It enumerates candidate files with `find` rather than listing them, so the two deleted scripts are simply not scanned. It still passes, and it still fails on a new producer.
- Two comments in `build-pam-module.sh` cited the deleted files as present tense ("both release scripts" in the header; `build-simple.sh` in the arch-check rationale). Both now say the script was deleted, with #241 alongside #222 — the historical explanation is worth keeping, a pointer to a nonexistent file is not.
- The CHANGELOG's two `### Fixed` bullets claimed those scripts were *fixed* to call the new builder, which is no longer what happened. They now say the two paths were removed outright by #241, and that what keeps the arch property is `build-pam-module.sh`'s machine-type check plus the release workflow stating an architecture for each build.

Verified after the rebase: `bash -n` on both new scripts, `./scripts/check-pam-module-producers.sh` (OK — one producer, and it verifies), `./scripts/check-cgo-quarantine.sh` (OK), `go build ./...`, `go vet ./...`, `make -n release`, and no reference to either deleted script remains anywhere in `scripts/`, `Makefile`, `.github/`, `CONTRIBUTING.md` or `DEPLOYMENT.md`. `ci.yml` auto-merged cleanly — #248's write-guard step and this PR's producer-check step are separate steps in the `Validate` job.

Issue #217, opened 2026-08-14
HIGH: a late device-flow completion resurrects an expired or revoked session into an active one


scttfrdmn:
Found by the v0.5.0 security audit. Verified at `d8ca4b3`.

An expired or explicitly revoked session can be **resurrected into an active one**
by a late device-flow completion.

## Evidence

`pkg/auth/broker.go:1203`:

```go
b.setSession(&updated)
```

`updated` is built from the stored session with `IsActive: true` and a fresh
`ExpiresAt`. There is **no re-check** of the session's current state before the
write: not `IsActive`, not `ExpiresAt`, not whether the session was removed by
`expireSessions` and re-created, and not whether an operator revoked it.

The device-flow completion path holds no lock across the read-modify-write either,
so it also races `expireSessions` and `RefreshSession` (see the separate refresh
issue) — last writer wins, and this one writes `IsActive: true`.

## Failure scenario

1. User starts a login. Session `S` is created pending (`IsActive: false`).
2. The user does not complete the device flow. `token_lifetime` elapses, or an
   operator runs a revocation, and `S` is marked inactive/expired.
3. The user's browser tab is still open. They complete the device approval **after**
   step 2. The provider returns tokens, the completion handler builds `updated` from
   `S`, sets `IsActive: true` and a fresh `ExpiresAt`, and writes it.

`S` is now a fully active session that outlives its own expiry and survives the
revocation. `CheckSession` will report it healthy, and the SSH provisioning path
will install a key for it.

The revocation case is the sharper one: an operator who revokes a session during an
in-progress login has no way to know their revocation will be undone seconds later.

## Fix

Re-read and validate under the same lock as the write:

```go
current, ok := b.getSession(id)
if !ok || current.Revoked || current.ExpiresAt.Before(time.Now()) {
    return b.createErrorResponse("SESSION_EXPIRED", ...), nil
}
```

and make the read-modify-write atomic rather than three separate operations. The
broker's session map already has a mutex; the completion path needs to hold it
across the whole sequence, or the store needs a compare-and-swap primitive.

## Test

Unit-testable with no Linux host: create a pending session, expire it, then drive
the completion path and assert the session is **not** active afterwards.

Issue #159, opened 2026-08-14
Identity binding accepts an email claim's local part, and no local account is off-limits: an IdP identity of root@anything can log in as root


scttfrdmn:
## What's wrong

`verifyIdentityBinding` accepts the **local part** of an email/UPN claim as a match
for the requested local account, and there is no denylist of privileged local
accounts anywhere in the tree.

`pkg/auth/broker.go:1356`:

```go
// If the claim is an email (e.g. preferred_username/email), allow matching
// either the full value or its local-part against the requested user.
if mapped == requested {
    return nil
}
if local, _, found := strings.Cut(mapped, "@"); found && local == requested {
    return nil
}
```

The local-part branch is reachable in every shipped configuration:

- `username_claim: "email"` — `configs/examples/broker.yaml:23`,
  `configs/production/broker-minimal.yaml:34`,
  `configs/production/broker-enterprise.yaml:29/50/86`, `broker-cloud.yaml:34`, and
  the hard-coded default in `loadFromEnvironment`.
- `username_claim: "preferred_username"` — `configs/providers/azure-ad.yaml`,
  `okta.yaml`. On Entra ID `preferred_username` **is** the UPN, i.e. an email, so the
  branch fires there too.

## Impact

Given any IdP-side ability to choose the local part of that claim — self-service
profile/alias editing (available in many tenants), a second verified domain, or a
guest/B2B/federated identity from another tenant — an attacker sets it to
`root@their.tld`, runs `ssh root@host`, approves the device flow with their own
credentials, and `verifyIdentityBinding` returns `nil`. `auth sufficient
pam_oidc.so` (`configs/pam/ssh:15`) then short-circuits the auth stack.
`configs/pam/su:11-13` gives the same for `su`.

`validateUserID` (`internal/ipc/validate.go`) accepts `root` without comment. There
is **no allowlist of OIDC-reachable local accounts and no denylist of `root` or
system accounts** anywhere in the repository.

The weaker variant needs no IdP control at all: guest `alice@partner.com` and
employee `alice@company.com` both bind to local `alice`.

## Why existing defences don't stop it

- This behaviour is deliberate and **tested as correct** —
  `pkg/auth/identity_binding_test.go:38-43` asserts `email: alice@example.com` +
  requested `alice` ⇒ no error. It will survive a casual refactor.
- The mismatch test at `:52-57` only proves `attacker != root`, not
  `root@evil.tld != root`.
[...]

Issue #219, opened 2026-08-14
HIGH: the device code reaches the logs at debug level and is fed to TokenFingerprint — journal readers can complete another user's login


scttfrdmn:
Found by the v0.5.0 security audit (two independent auditors flagged the
`TokenFingerprint` half). Verified at `d8ca4b3`.

The device code — the bearer credential that completes a login — reaches the logs
in two places. Under the shipped systemd unit both land in the journal, which is
group-readable by `systemd-journal` on Debian/Ubuntu.

## Evidence

**1. Logged at debug in the auth path.** The device code (and/or the full
verification URI with the code embedded) is passed to a debug-level log call in
the device-flow handling in `pkg/auth`. `configs/examples/broker.yaml` and the
troubleshooting instructions in `DEPLOYMENT.md` both tell operators to set
`log_level: debug` when a login is failing — which is exactly when a device code
is in flight.

**2. `TokenFingerprint` is fed a device code.** The fingerprint helper exists to
produce a non-reversible short identifier for correlating log lines about a token.
It is called with the device code as its input, so the fingerprint of a *live
credential* is what gets recorded. That is not itself a disclosure, but it means
the code is present in the caller's frame and defeats the purpose of having a
fingerprint helper at all: the helper's contract is "this value is safe to log
because it is not the secret", and here the secret and the fingerprinted value are
the same class of thing.

## Failure scenario

An operator is debugging a failing login and sets `log_level: debug`, as the
project's own docs instruct. A user starts a login.

The device code is now in the journal. On Debian/Ubuntu, any account in
`systemd-journal` can read it — and so can anything that ships the journal
off-host, which is normal practice. The device-flow window is minutes, and the code
is all that is needed to complete the authentication as that user, from anywhere.
The legitimate user's own login then either succeeds (and they see nothing wrong) or
fails with a used-code error they will read as a glitch.

The attacker needs no OIDC account of their own.

## Fix

1. Never log the device code, the `verification_uri_complete`, or any value derived
   directly from them — not at debug, not at trace. Log the `session_id` instead,
   which is already the correlation key everywhere else.
2. Stop passing the device code to `TokenFingerprint`. If a correlation handle for
   the device flow is wanted, fingerprint the session ID.
3. Add a test that asserts a device-flow log capture at debug level contains
   neither the code nor the complete URI. A grep-based unit test over a captured
   log buffer is enough and will catch this regressing.

Related: `#209` (config file mode 0644) is the same class of problem in a different
[...]

PR #253, opened 2026-08-14
fix(broker): stop an abandoned device flow from denying the account its login

### LINKED ISSUE #163: Pending-session accounting lets an unauthenticated remote client deny logins to a chosen user, and to the whole host
## What's wrong

`countUserSessions` counts **pending** sessions toward the per-user
`MaxConcurrentSessions` cap (default 10, `pkg/config/config.go:419`). The session is
inserted at `pkg/auth/broker.go:409` — before the cap at `:304-315` can help — and is
only removed when the device code expires (`:831-834`), a lifetime the *provider*
chooses and the broker permits up to `maxDeviceCodeExpiry = 86400`
(`device_flow.go:322-325`).

## Impact

Unauthenticated remote attacker, two variants:

**Targeted.** Ten abandoned SSH attempts against `victim`
(`for i in $(seq 10); do ssh victim@host & done`, then disconnect) make the eleventh
return `TOO_MANY_SESSIONS` → `PAM_PERM_DENIED` for the whole device-code lifetime.
Refreshing the attempts sustains it indefinitely. **The victim cannot log in.**

**Host-wide.** Ten usernames × ten attempts reaches `maxPendingFlows = 100`
(`:412-421`) and every login on the host gets `RATE_LIMITED`.

Each attempt also costs one authenticated POST to the IdP's device endpoint using the
broker's client credentials (`:360`, synchronous, before any of the caps) — free
amplification against the IdP's own rate limits, from an unauthenticated attacker.

## Why existing defences don't stop it

- The per-user cap is the mechanism being abused, not a defence against it.
- `Authenticate` never checks the requested `UserID` is a real local account, so
  nonexistent-but-syntactically-valid names work just as well.
- The IPC limiter is the always-uid-0 global bucket (see the companion issue).
- `maxPendingFlows` is checked *after* `setSession`, so even a rejected flow
  transiently occupies a session slot.

## Fix

1. Do not count pending sessions toward `MaxConcurrentSessions`; give pending flows a
   smaller, separately configured cap keyed on user **and** source IP.
2. Bound a pending session's own lifetime independently of the provider's
   `expires_in`, e.g. `min(expires_in, 15m)`.
3. Resolve the local account (`user.Lookup`) **before** starting a device flow, so a
   request for a nonexistent user costs nothing and never reaches the IdP.

## Acceptance criteria

- A test asserting 10 abandoned pending flows for user A do not prevent A's next
  login once they are older than the pending cap's window.
- A test asserting an `authenticate` for a nonexistent local user performs no
  outbound device-authorization request.

scttfrdmn:
## What was wrong

The broker counted a login nobody had approved as though it had succeeded. An authenticate request names the account it wants, sshd runs its PAM stack for whatever username a client offers before that client has authenticated anything, and every such request became a session in the broker's map that counted towards the named account's `max_concurrent_sessions`. Ten connection attempts naming one account — none of them completed, which is what makes the attack free — filled that account's limit with authentications nobody had approved, and the real user's next login was refused `TOO_MANY_SESSIONS`. The entries went away when the device code the provider had issued expired, a lifetime the provider chooses and may set as high as 24 hours.

The host-wide cap that was supposed to bound this was the second half of the same problem. A hundred unfinished flows, spread across invented usernames so no single account limit was reached, made every login on the machine answer `RATE_LIMITED` until the codes expired. It was also checked *after* the session had been stored and after the round trip to the identity provider, so a request that was about to be refused had already spent an authenticated POST to the provider's device endpoint and briefly occupied a session slot on the way out.

## Reachability, stated precisely

The broker's IPC socket is root-owned, `0660`, and `verifyPeerCredentials` accepts uid 0 only, so the peer is always root. That does **not** make this a root-only defect, and the issue's "unauthenticated remote client" framing is essentially correct: the peer is sshd's PAM stack, which runs as root *before* the remote client has authenticated and forwards that client's chosen username — deliberately, so that a valid and an invalid user cost the same time. The requests therefore come from a client that has proved nothing, over ssh. The one field it does not control is `source_ip`, which the module derives from `PAM_RHOST`; that is what makes the address usable as a second key.

Severity is denial of service, not privilege escalation: nothing here grants access. It can be aimed at one chosen account or at the whole host.

## Claims verified

Line numbers in the issue are stale; the substance holds. Against `main` at the time of this branch:

| Claim | Where | Verdict |
|---|---|---|
| Pending sessions count towards `max_concurrent_sessions` | `pkg/auth/broker.go:554` (`countUserSessions`, no `IsActive` test) | Confirmed |
| The session is stored before the host-wide cap is checked | the `b.setSession(session)` / `atomic.AddInt64(&b.pendingFlows, 1)` pair this diff removes | Confirmed |
| A pending session's removal is driven by the provider's device-code expiry | `ExpiresAt` came from `sessionExpiry`, the poll loop's timer from `deviceFlow.ExpiresAt`, bounded only by `maxDeviceCodeExpiry` = 86400 | Confirmed |
| Nothing checks that the requested account exists before the IdP is contacted | `verifyIdentityBinding` runs at the end of the flow | Confirmed |
| A full pool refuses everyone | the old cap answered `RATE_LIMITED` to whichever request arrived next | Confirmed |

## What changed

Unfinished logins are accounted for separately from sessions. `max_concurrent_sessions` counts only logins that completed — what a limit on how much access one identity holds should count. A login still awaiting device approval is charged instead against the `(account, source address)` pair that asked for it: the account is the one field a remote client chooses freely, so a cap keyed on it alone can always be filled on somebody else's behalf, while the address is where the connection came from. An attacker can exhaust the budget of `(victim, its own address)` and nothing else.

The host-wide pool is still bounded at 100, but reaching it no longer refuses the login that arrives next: it displaces the oldest pending flow of whichever pair holds the most, and only a pair holding strictly more than the requester can be displaced. Holding the largest share is exactly what selects you to pay for it, so hoarding cannot lock the host out, and a login from a pair holding nothing is always served. That is the argument the IPC limiter's LRU bucket eviction already rests on. Only when nobody holds more than the requester — a pool genuinely full of logins with as good a claim — is `RATE_LIMITED` returned.

An unfinished login now expires on the broker's clock (`pending_auth_lifetime`, 15m default) rather than on the provider's `expires_in`, and that deadline also bounds the goroutine polling for it. Removing a pending session — by revocation, by the sweep, or by displacement — closes a per-session cancellation channel that stops its poll loop. Without it a cap on how many unfinished logins the host holds bounded none of the work each one costs: the goroutine kept a live device code warm for up to the code's full lifetime after its session was gone. Abandoning a flow that way is audited with the same `SESSION_GONE` code the completion path already uses for the same condition (#217), so an operator who revokes a session mid-flow still gets a record of the login it interrupted — and now nothing was minted in the meantime for that record to withdraw. Admission also decides *before* `StartDeviceFlow`, so a refused login costs the identity provider nothing, and the slot is returned immediately if the provider errors.

Finally, a login naming an account this host does not have is refused before the provider is contacted (`require_local_account`, default true). Such a login could never have succeeded — sshd needs a passwd entry once the stack returns, and the identity binding resolves the account again at the end of the flow — so all that changes is what it costs to ask, which until now was an authenticated request to the site's IdP and a slot in the pending pool, available to every ssh scanner on the internet. The trade is an enumeration oracle (a client can tell a real account from an invented one by whether it is offered a device code), which is why the setting exists to turn it off. The gate does **not** treat an NSS *error* as "no such account" — refusing every login while `getpwnam` is briefly unavailable would be the same denial of service by another route; the privileged-account guard on the completion path is the one that fails closed there (#159).

Two smaller things fall out. A pending session no longer stores the device code in `TokenFingerprint` (nothing read it, and the completion path overwrites it with the real fingerprint), and the `expires_at` handed to the client is now the earlier of the provider's expiry and the broker's window, rather than promising a window the broker will not honour.

### Settings

Three keys, in `pkg/config` with the other authentication settings, defaulted via `setDefaults` so they apply to files written before they existed, clamped at both ends — an oversized value saturates rather than being honoured, the fail-open shape `clampToInt32` exists for in the IPC limiter — and documented in `configs/CONFIGURATION-GUIDE.md`:

| Key | Default | Bounds |
|---|---|---|
| `authentication.max_pending_auths_per_source` | `5` | 1–32, `0` means the default |
| `authentication.pending_auth_lifetime` | `15m` | 1m–1h, `0` means the default |
| `authentication.require_local_account` | `true` | — |

## Tests

`pkg/auth/pending_sessions_test.go` drives everything through `Authenticate`, since these were properties of how counting and storing were ordered rather than of any one helper. Its device endpoint counts requests, so "a refused login must not have reached the provider" is an assertion rather than a claim. No test sleeps: pending entries are aged by rewriting the stored session's timestamps, and the one wait is a 10s *guard* on a goroutine that has already exited when the code is correct.

The regression gate is `TestAbandonedPendingFlowsDoNotDenyTheAccountItsLogin`: ten abandoned requests for `alice` from one address against `max_concurrent_sessions: 3`, then `alice`'s own login from a different address, which must be served.

### Non-vacuity proof

For each guard: back up `pkg/auth/broker.go`, perturb with `perl -0pi -e`, run *only* the covering test, restore, `diff -q`. All six perturbations produced a failure, and every restore was byte-identical.

| Guard perturbed | Covering test | Failure observed |
|---|---|---|
| `&& session.IsActive` dropped from `countActiveUserSessions` | `TestAbandonedPendingFlowsDoNotDenyTheAccountItsLogin` | `attempt 4 was refused "TOO_MANY_SESSIONS"` |
| per-source cap (`if false && mine >= perSourceCap`) | same | `10 of 10 abandoned logins were admitted, want 5` |
| deadline skip in admission counting | `TestPendingLoginSlotIsReleasedWhenItsWindowCloses` | refused `TOO_MANY_PENDING_AUTHS` after every entry was past its deadline |
| pending expiry hardcoded to 24h | `TestPendingLifetimeIsBoundedByTheBrokerNotTheProvider` | `expires ... about 24h0m0s from now` |
| displacement disabled (`if true \|\| count <= mine`) | `TestFullPendingPoolDisplacesItsLargestHolder` | a login from a pair holding nothing was refused `RATE_LIMITED` |
| `close(session.cancel)` disabled | `TestRevokingAPendingLoginStopsItsPolling` | guard fired: the poll goroutine was still running |

`TestDeviceFlowWillNotActivateASessionThatIsGone` (the #217 gate) needed reworking rather than removing: a revocation now cancels the flow early, so the test's setup no longer reached `replaceSession` at all. It now simulates the ordering that *is* still reachable and is the one that can leave a credential behind — the revocation landed while the goroutine was inside the token request, past the select that watches for cancellation — so the pointer-identity CAS is still exercised, along with the key and token withdrawal.

`gofmt -l .` empty, `go build ./...`, `golangci-lint run` on both packages clean, `go test ./...` green, `go test -race ./pkg/auth ./pkg/config` green, and `-race -count=3` over the touched tests green. (The macOS cgo failures for the `//go:build linux` files are absent here because nothing in this change touches them.)

## Found, not fixed

- A login with no source address — the console, or a `PAM_RHOST` that is a hostname rather than an address (#169) — shares one pending budget per account. Those cannot be produced remotely, which is what makes the sharing acceptable, but it is a sharing.
- Two clients behind one NAT address share a budget for the same account, so an attacker who is on the victim's network can still consume that pair's five slots. Closing that needs something the client cannot choose and does not share, which at this point in the flow does not exist.
- A pending session's idle timeout still runs from its creation (`CheckSession` deliberately does not touch `LastAccessed` while pending), so an `idle_timeout` shorter than `pending_auth_lifetime` sweeps unfinished logins early. Harmless — it ends logins nobody is completing — but it means two settings can bound the same window.
- `RateLimiter`'s doc comment in `internal/ipc/ratelimit.go` still describes this pending-flow accounting as future work. Another package, and another agent's files this week, so it is left alone.

Closes #163

### DIFF
--- a/configs/CONFIGURATION-GUIDE.md
+++ b/configs/CONFIGURATION-GUIDE.md
@@ -566,6 +566,9 @@ absent.
 | `server.require_peer_auth` | `true` | Verify the peer's credentials over `SO_PEERCRED` and accept uid 0 only. Turning this off lets any local process talk to the broker; there is no reason to. |
 | `server.metrics_addr` | *(unset)* | TCP address for the Prometheus `/metrics` endpoint, e.g. `127.0.0.1:9090`. Unset means no listener. The endpoint is unauthenticated — bind it to loopback. |
 | `authentication.idle_timeout` | *(unset)* | Expire a session this long after its last use, in addition to `token_lifetime`. Unset means only `token_lifetime` applies. |
+| `authentication.max_pending_auths_per_source` | `5` | How many logins one account may have awaiting device approval at once from one source address. Clamped to 1–32; `0` means the default, not "unlimited". See below. |
+| `authentication.pending_auth_lifetime` | `15m` | How long a login that never completes keeps its slot. Clamped to 1m–1h. See below. |
+| `authentication.require_local_account` | `true` | Refuse a login for an account this host does not have before contacting the provider. See below. |
 | `authentication.allow_privileged_accounts` | *(empty)* | The named exceptions to "no OIDC identity may log in as uid < 1000". See above. |
 | `authentication.geoip_database_path` | *(unset)* | MaxMind GeoLite2 database for `geo_restrictions`. Without it every country code is empty, so an `allowed_countries` restriction denies everyone. |
 | `authentication.location_history.*` | 90d, 10 entries, memory | Window, per-user cap and `persist_path` for the "unusual location" risk signal. |
@@ -575,6 +578,37 @@ absent.
 | `audit.overflow_strategy` | `block` | What happens when that channel is full: `block` (backpressure), `sync` (write inline), or `drop` (discard and count). `drop` loses audit records and must be chosen deliberately. |
 | `audit.outputs[].type` | — | `file`, `stdout`, `syslog` or `http`. Any other value is refused at startup. |
 | `audit.enabled` | `true` | Whether to audit at all. With no `audit.outputs` alongside it the broker refuses to start, rather than accepting events and discarding them (#210). |
+
+### Logins Nobody Has Completed Yet
+
+Between showing a verification URL and the user approving it, the broker is
+holding a login that has proved nothing about who asked for it: sshd runs the PAM
+stack for whatever username a remote client offers, before that client has
+authenticated. The three settings above bound what such a login may cost.
+
+`max_pending_auths_per_source` is counted per (account, source address) pair, not
+per account. The account is the part of the request a client chooses freely, so a
+cap on it alone can be filled on somebody else's behalf — which is what #163
+was: unfinished logins counted towards `max_concurrent_sessions`, so a handful of
+connection attempts naming one account locked its owner out. They no longer count
+towards it. Raise this if your users routinely open more than five terminals at
+once and see `TOO_MANY_PENDING_AUTHS`; the value is clamped to 32 because a
+per-source cap as large as the host-wide pool is not a cap.
+
+`pending_auth_lifetime` is the broker's own deadline on an unfinished login,
+replacing the device code's `expires_in` — a number the provider chooses, and may
+set as high as 24 hours. The PAM module gives up after 90 seconds, so anything
+still pending after that is a login nobody is waiting for. Shortening it below a
+minute is not possible; a user who has to fetch their phone needs the window.
+
+`require_local_account` refuses a login naming an account this host does not have
+before the provider is contacted. Such a login could never have succeeded — sshd
+needs a passwd entry, and the identity binding resolves the account again at the
+end of the flow — so all this changes is what it costs to ask: without it, every
+ssh scanner on the internet starts device authorizations against your identity
+provider. The trade is that a remote client can tell a real local account from an
+invented one by whether it is offered a device code. Set it to `false` if that
+matters more to you than the requests.
 
 ## Authentication Policies
 
--- a/pkg/auth/broker.go
+++ b/pkg/auth/broker.go
@@ -13,7 +13,6 @@ import (
 	"strconv"
 	"strings"
 	"sync"
-	"sync/atomic"
 	"time"
 
 	"github.com/rs/zerolog/log"
@@ -37,7 +36,6 @@ type Broker struct {
 	stopChan              chan struct{}
 	wg                    sync.WaitGroup
 	metrics               *oidcmetrics.Metrics // nil when metrics are disabled
-	pendingFlows          int64                // atomic counter of in-progress device authorization goroutines
 	version               string               // build version, set via SetVersion; reported by Status
 	startedAt             time.Time            // when Start ran; zero until then. Reported by Status
 
@@ -118,6 +116,19 @@ type Session struct {
 	RequireDeviceTrust bool
 
 	Metadata map[string]interface{}
+
+	// cancel is closed when a session that has not completed its device flow is
+	// removed from the store, which is how the goroutine polling for that
+	// authorization learns to stop. It is nil once the flow has completed, and nil on
+	// a session built by anything other than Authenticate — a nil channel simply
+	// never fires in the poll loop's select.
+	//
+	// (#163) Without it, a pending session and its polling goroutine had independent
+	// lifetimes: revoking, sweeping or displacing the session left the goroutine
+	// holding a live device code and polling the provider for it until the code's own
+	// expiry, up to 24 hours later. A cap on how many pending logins the host holds
+	// means nothing if the work each one costs outlives the accounting.
+	cancel chan struct{}
 }
 
 // AuthRequest represents an authentication request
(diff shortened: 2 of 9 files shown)


scttfrdmn:
Rebased onto \`main\` by hand — \`update-branch --rebase\` hit a \`RebaseConflictError\`.

One conflict, in \`pkg/auth/broker.go\`: this PR moves the pending-session construction to *before* \`StartDeviceFlow\`, so the old block (session literal, \`setSession\`, and the \`maxPendingFlows = 100\` goroutine cap) is deleted here, while #249 had edited that same block on \`main\`. Resolved in favour of this PR — the \`b.pendingFlows\` counter and its cap are superseded by \`admitPendingSession\`, and the field itself is gone, so keeping either side of the old block would not compile.

#249's substance is preserved: the new session literal never sets \`TokenFingerprint\`, which is exactly the #219 fix, and I carried the comment explaining why across so the reason does not vanish with the block. \`device_code_secrecy_test.go\` still asserts it and still passes.

\`go build\`, \`go vet\`, \`gofmt\`, and \`go test ./... -count=1\` (plus \`./pkg/auth -race\`) all clean locally.

Issue #221, opened 2026-08-14
MEDIUM: upload-artifact@v7 and download-artifact@v8 are the only unpinned actions in the repo, and they sit in the release artifact path


scttfrdmn:
Found by the v0.5.0 security audit. Verified at `d8ca4b3`.

Two unpinned actions in the workflow that **builds and publishes the release
artifacts** — the one workflow where a compromised action is worst.

## Evidence

Every other action in the repo is pinned to a commit SHA with a version comment,
e.g. `release.yml:17`:

```yaml
uses: actions/checkout@9c091bb21b7c1c1d1991bb908d89e4e9dddfe3e0 # v7.0.0
```

The two exceptions are both in `release.yml`, and both are in the artifact path:

- `:148` — `uses: actions/upload-artifact@v7`
- `:166` — `uses: actions/download-artifact@v8`

Verified: these are the **only** unpinned `uses:` lines in the repository.

## Failure scenario

`v7`/`v8` are mutable tags. Whoever controls the `actions/upload-artifact`
repository — or anyone who compromises it — can retarget the tag.

The compromised step sits exactly between "build the PAM module and broker" and
"publish them to the releases page". `upload-artifact` receives the built
`pam_oidc.so`, `oidc-auth-broker` and the tarball; `download-artifact` produces
what `gh release create` uploads. A malicious version of either can substitute the
binaries, and the published `.sha256` files are generated **before** the upload
(`:145`), so they would be recomputed against — no, worse: they are computed on the
correct artifacts and uploaded alongside, so a substitution at download time yields
a checksum mismatch, but nothing in the workflow *verifies* the checksums after
download. Nothing detects it.

The artifacts are a root-run broker and a PAM module that
`scripts/install-release.sh` installs into the auth stack of every consumer host.

## Fix

1. Pin both to commit SHAs with version comments, matching the convention already
   used everywhere else.
2. Verify the `.sha256` files in the `release` job after download, before
   `gh release create`. The files are already produced; nothing checks them.
3. This is the pinning half of #180 (cosign + SLSA) — signing the artifacts is
   worth little if the build can swap them before signing. Do the pinning first;
   it is a two-line change.

## Test

Add a `Validate`-job check that greps every `uses:` in `.github/workflows/` for a
40-hex SHA and fails on a bare tag. That makes this class of regression impossible
rather than merely fixed once.

PR #260, opened 2026-09-04
fix(security): restore Trivy scanning and triage every open code-scanning alert

### LINKED ISSUE #258: SECURITY TOOLING: ten Trivy alerts including a critical are orphaned on main, and nothing can ever close them
## What is wrong

The Security tab shows **1 critical and 3 high** dependency alerts on `main`. All four are false, and so are the six others alongside them. The critical one reads:

```
Package: golang.org/x/crypto
Installed Version: v0.14.0
Vulnerability CVE-2024-45337
Severity: CRITICAL
Fixed Version: 0.31.0
```

`go.mod` has `golang.org/x/crypto v0.55.0`. Three of the ten name `github.com/go-jose/go-jose/v3`, which is not a dependency at all any more — only `v4` is.

All ten were created `2025-12-21` and have **never been updated since**:

| CVE | package Trivy saw | actually in go.mod |
|---|---|---|
| CVE-2024-45337 (critical) | x/crypto v0.14.0 | v0.55.0 |
| CVE-2025-22869 (high) | x/crypto | v0.55.0 |
| CVE-2024-24786 (high) | protobuf | v1.36.11 |
| CVE-2025-22868 (high) | x/oauth2 | v0.36.0 |
| CVE-2023-48795, CVE-2025-47914, CVE-2025-58181 | x/crypto | v0.55.0 |
| CVE-2024-28180, CVE-2025-27144, GHSA-2c7c-3mj9-8fqh | go-jose **v3** | only v4 present |

## Why they cannot be closed

Because Trivy is not in CI. `grep -rin trivy .github/` finds nothing, and `git log -S trivy -- .github/` puts its removal in `9b6c0ae` ("Simplify to minimal CI for solo development").

GitHub retires a code-scanning alert only when the same tool re-runs and stops reporting it. With the tool gone, the alerts it had already uploaded are immortal: no future run will ever contradict them.

## Why this matters beyond tidiness

Two ways, and the second is the one that bites:

1. A critical badge that is false teaches everyone to discount the Security tab, including the parts that are true. There are 54 open alerts; the loudest four are fiction.
2. **The scanning gap is real.** Trivy is the only tool here that scanned the container image and the full transitive module tree in one pass. It is not running, so a genuine finding of the kind it catches would not appear either — and the stale alerts make the tab *look* like it is being watched. This is the same defect shape as #210 and #212: a control that reports something other than what it does.

Nothing corroborates the ten: Dependabot reports 0 open alerts, and the `Go Vulnerability Check` (govulncheck) job passes on `main`.

## Fix

Restore Trivy to `.github/workflows/security.yml` — filesystem scan, SARIF output, `upload-sarif` with its own category, action pinned by SHA like every other action in that file (#221). One run then reconciles all ten against the real dependency tree and closes them, and the scanning gap closes with them.

Dismissing the ten by hand is the alternative and is worse: it clears the badge while leaving nothing scanning, which is the state that produced this issue.

### LINKED ISSUE #259: SECURITY TOOLING: 23 gosec and 4 CodeQL alerts sit open on main, untriaged and unannotated — and one of them is a real bypass
## What is wrong

27 of the 54 open code-scanning alerts on `main` come from gosec (23) and CodeQL (4). None has ever been triaged, and the repository contains **no `#nosec` annotation anywhere** — so there is no record of which are accepted and why. `3f12a5f` already did this work for `nosemgrep`; the gosec equivalent never followed.

Triaged now, they are three different things:

### 1. G123 — a real bypass of configured certificate pinning

`pkg/auth/device_flow.go:313` sets `tls.Config.VerifyPeerCertificate` to enforce `security.tls_verification.pinned_certificates`, and sets no `VerifyConnection`.

**`VerifyPeerCertificate` is not called on a resumed TLS session.** `VerifyConnection` is. So on any handshake that resumes a cached session, the pin check does not run and the connection is accepted on the strength of ordinary chain verification alone — while the operator has been told, by the config key they set, that only their pinned certificates are trusted.

That is the same shape as #210 and #212: a control that is configured, reported as active, and not applied on some path. The broker makes repeated calls to the IdP per login (discovery, device authorization, token, then polling), which is exactly the traffic pattern resumption exists for.

Fix: perform the pin check in `VerifyConnection`, which is invoked on both full and resumed handshakes and receives the peer chain in `ConnectionState.PeerCertificates`, and prove it with a test that resumes a session and asserts the check still runs.

### 2. G306 — a public key written 0644

`pkg/ssh/keys.go:208` writes `id_rsa.pub` with `0644` inside a `0700` directory. Nothing outside the broker reads it — `LoadKey` is the only reader, in-process. Not exploitable, but the permission is wider than any purpose it serves. Tighten to `0600`.

### 3. Twenty-one accepted findings with no record that they are accepted

- **5 × G304** (file inclusion via variable): `pkg/ssh/keys.go:240,247,254` — path is `filepath.Join(baseDir, keyID)` where `keyID` has already passed `validateKeyID`, an allowlist of `^[A-Za-z0-9][A-Za-z0-9_-]{0,127}$`, so it can contain neither `/` nor `.` and cannot traverse. `pkg/config/config.go:905` — the `file:` secret indirection, reading a path the operator wrote in their own root-owned config; a config that can name a file can equally name the secret inline, so this grants nothing. `pkg/auth/location_history.go:227` — the broker's own state path.
- **16 × G103** (use of unsafe calls): all in `cmd/pam-module/{auth,args,recv}_linux.go`, crossing the cgo boundary. Since #198 the shipped module is pure C; these Go files exist only to let `go test` drive it. `unsafe.Pointer` is not avoidable there.

These want `#nosec Gxxx -- <reason>` at each site, naming the gate that makes them safe. Not a blanket `-exclude=G103`, which would also silence a genuinely dangerous `unsafe` use somewhere it matters.

### 4. CodeQL log-injection ×4

`test/e2e/fakeoidc/main.go:185` (×3) and `:210` — user-controlled values logged unsanitised in the fake IdP. Not shipped and not privileged, so the impact is nil, but the alerts are permanent noise: the gosec job passes `-exclude-dir=test` and CodeQL has no equivalent exclusion. Sanitise the logged values.

## Why file this rather than just dismiss

The G123 finding is real and would have stayed invisible under a blanket dismissal — which is the argument for triaging the whole set once, in one place, with the reasons written down where the next reader will find them.

scttfrdmn:
Closes #258, closes #259.

`main` carried **37 open code-scanning alerts**, none triaged: 10 Trivy, 23 gosec,
4 CodeQL. One was a real defect; the rest were noise that nothing could ever
clear. Both are worth fixing, because a Security tab that is 97% permanent noise
is one nobody reads the other 3% of.

## Trivy — #258

Every security workflow was deleted in `9b6c0ae` (Dec 2025). Code scanning
retires an alert only when the tool that raised it runs again and stops reporting
it, so removing a scanner doesn't clear its findings — it makes them **immortal**.
The ten alerts, one critical, then described dependency versions this repo no
longer has for eight months; three name `go-jose/v3`, which is in neither
`go.mod` nor `go.sum`.

Restored with the action pinned by SHA (#221). **Two details are load-bearing:**

- the job id stays `trivy`, and
- the upload passes **no `category:`**.

GitHub derives the analysis key from workflow path + job id; the orphans carry
`.github/workflows/security.yml:trivy` with no category — read off
`most_recent_instance.analysis_key`, not assumed — and retirement is
per-analysis-key. **The explicit category I proposed in #258 would have created a
new key and left all ten unclosable forever.** That's a correction to the issue,
not a deviation from it.

Run against this tree locally, the restored scan reports **none of the ten**, so
the first run on `main` reconciles and closes them.

It does report two things:

| Finding | Disposition |
|---|---|
| `CVE-2026-56855`, `CVE-2026-78662` — x/crypto ssh mux DoS, fixed in 0.56.0 | Filed separately: 0.56.0 requires `go 1.26.0`, i.e. a toolchain bump across every workflow. Neither is reachable — govulncheck is green. Not smuggled into a scanning PR. |
| `GO-2026-5932` — "x/crypto/openpgp is unmaintained", no fixed version | New `.trivyignore.yaml`, reason recorded in the repo rather than dismissed in the UI. `go list -deps ./...` does not name openpgp. |

`scanners: vuln,secret` (Trivy's filesystem default, written out as a decision);
`misconfig` stays off — the only files it has to look at are three test-harness
Dockerfiles whose containers run as root deliberately.

## gosec — #259, one real finding

**G123: certificate pinning was enforced in `VerifyPeerCertificate`, which a
resumed TLS session never calls.** From the second handshake onwards a pinned
configuration would have stopped checking its pins.

It was **not reachable** as the code stood: client-side resumption needs a
non-nil `tls.Config.ClientSessionCache`, `buildTLSConfig` sets none, and
`net/http` does not add one to a caller-supplied `TLSClientConfig`. So the pins
were enforced — but only by that absence, which no comment stated and no test
held. Adding a session cache anywhere in the request path would have silently
disabled pinning.

The check moves to `VerifyConnection`, which runs on every handshake and *does*
receive the peer chain on a resumed one. The new test installs a cache, drives
three requests, and fails unless a handshake resumed **and** the pin check ran on
all three. Reverting the fix makes it fail — which is the only reason to believe
it.

**G306:** the public half of a provisioned key pair was `0644`; its only reader
is `LoadKey` in the same process, and the directory is already `0700`.

**The other 21 are annotated, not excluded.** Five `G304` reads take a path from
root-owned config or from an identifier that already passed `validateKeyID`,
whose allowlist admits neither `/` nor `.` — recorded together with the note that
the annotations must come off if that pattern is loosened. Sixteen `G103` sites
are the two shapes cgo leaves no alternative to, in test wrappers that have
shipped in nothing since #198. No rule excluded, no directory skipped.

Verified in the configuration CI actually uses (golang:1.25 container):
**41 files checked including all six `_linux.go` files, 31 nosec (was 15),
0 issues.**

That run also explained *why* the gosec job covers `cmd/pam-module`: it doesn't
install the PAM headers, cgo fails, and `go/packages` falls back to the real
sources instead of handing gosec build-cache files. With the headers present:
6 findings at `/root/.cache/go-build` paths and the `_linux.go` files never
opened. So "fixing" that job by adding the dev packages would silently drop the
PAM module from the scan — the job now says so in a comment.

## CodeQL — #259

The e2e fake IdP logged request-controlled values unescaped, so a path or
`?username=` containing a newline could forge log lines in
`docker compose logs fakeoidc` — which is exactly what an engineer reads to find
out why a case failed. Test-harness only, unprivileged, shipped in no release,
but a log the thing under test can write is not evidence. Values go through
`strconv.Quote` with a length bound.

## Checks

- `gofmt` clean, `go build ./...`, `go vet ./...`, `go test ./...` — all pass
- `trivy fs` locally: 3 findings, all accounted for above
- gosec in a linux container, CI's configuration: 0 issues
- G123 fix mutation-tested: reverting it fails both assertions

### DIFF
--- a/.github/workflows/security.yml
+++ b/.github/workflows/security.yml
@@ -68,6 +68,22 @@ jobs:
           go-version: '1.25.13'
           cache: true
 
+      # This job deliberately does not install libpam0g-dev/libjson-c-dev, unlike the
+      # govulncheck job below, and that is what makes it scan cmd/pam-module at all.
+      #
+      # gosec loads packages through go/packages, which for a cgo package hands back
+      # the *generated* Go files rather than the sources. With the PAM headers present
+      # cgo succeeds, and gosec then reports against paths inside
+      # /root/.cache/go-build — findings that point at no file in this repository and
+      # cannot become a useful alert, while the 16 unsafe.Pointer sites in
+      # {auth,args,recv}_linux.go go unexamined. Without the headers cgo fails, the
+      # loader falls back to the real sources, and gosec analyses the code a reviewer
+      # can actually read. Measured both ways in a golang:1.25 container: with the
+      # headers, 6 findings at build-cache paths and the _linux.go files never opened;
+      # without them, 41 files checked including all six _linux.go files.
+      #
+      # So if this job is ever "fixed" by adding the dev packages, coverage of the PAM
+      # module silently disappears.
       - name: Run Gosec Security Scanner
         run: |
           # Install gosec v2.26.1 (Go 1.25 compatible)
@@ -225,3 +241,74 @@ jobs:
         uses: github/codeql-action/upload-sarif@60d8f0d1f1f8c8d07ef53bd027032705d414ec28 # v3
         with:
           sarif_file: semgrep-results.sarif
+
+  # Restored (#258). This job existed until 9b6c0ae removed every security
+  # workflow, and its ten alerts — one critical — stayed open on main for the eight
+  # months since, describing dependency versions this repository no longer has.
+  # Code scanning retires an alert only when the tool that raised it runs again and
+  # stops reporting it, so removing a scanner does not retire its findings: it
+  # makes them permanent.
+  #
+  # Two things about this job are load-bearing and must not be "tidied":
+  #
+  #   - The job id is `trivy` and this file is `.github/workflows/security.yml`.
+  #   - The upload step passes no `category:`.
+  #
+  # GitHub derives the analysis key an alert is filed under from the workflow path
+  # and the job id, and the ten orphans carry
+  # `.github/workflows/security.yml:trivy` with no category (checked, not assumed:
+  # `most_recent_instance.analysis_key` on the open alerts). Retirement is
+  # per-analysis-key, so renaming the job, moving the file, or adding a category
+  # produces a *new* key and leaves the orphans exactly as unclosable as they are
+  # now. #258 proposed giving this upload its own category; that would have
+  # defeated the point of the issue.
+  trivy:
+    name: Trivy Security Scan
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: read
+      security-events: write
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      # What this adds over the govulncheck job above: govulncheck reports a
+      # vulnerability only when it can reach the affected symbol from this code,
+      # which is the right signal for "fix this now" and the wrong one for "what is
+      # in our dependency tree". Trivy answers by version, so a dependency that is
+      # vulnerable but not currently called still shows up — the case that becomes a
+      # live one the moment somebody writes a new call.
+      #
+      # `scanners: vuln,secret` is Trivy's default for a filesystem scan, written out
+      # so that it is a decision rather than a default. `misconfig` is deliberately
+      # off: the only files it has to look at here are the three test-harness
+      # Dockerfiles, whose containers run as root on purpose (the broker under test
+      # needs root, and the images are built by `make e2e` and thrown away), so it
+      # reports six findings of which none is actionable. Turn it on if this project
+      # ever ships a container image.
+      - name: Run Trivy filesystem scan
+        uses: aquasecurity/trivy-action@ed142fd0673e97e23eac54620cfb913e5ce36c25 # v0.36.0
+        with:
+          scan-type: fs
+          scan-ref: .
+          scanners: vuln,secret
+          format: sarif
+          output: trivy-results.sarif
+          # Trivy does not pick .trivyignore.yaml up on its own — verified: the same
+          # scan reports GO-2026-5932 without this input and drops it with the input.
+          # Only the plain `.trivyignore` name is a default.
+          trivyignores: .trivyignore.yaml
+
+      # No `continue-on-error:` here, unlike the gosec and semgrep jobs. Those two
+      # tolerate a non-zero exit because their scanners exit non-zero *on findings*;
+      # Trivy with `format: sarif` does not (it needs `exit-code:` for that), so the
+      # only way this step fails is that the scan itself did not run — a DB download
+      # that 429'd, say. That has to be a red job. A scan that quietly did not happen,
+      # and an upload that therefore quietly did not reconcile anything, is precisely
+      # how ten alerts came to sit on main for eight months.
+      - name: Upload Trivy results to GitHub Security
+        uses: github/codeql-action/upload-sarif@b20883b0cd1f46c72ae0ba6d1090936928f9fa30 # v4.32.0
+        with:
+          sarif_file: trivy-results.sarif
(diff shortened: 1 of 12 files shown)


scttfrdmn:
Filed the deferred half as #261: the two x/crypto ssh CVEs need v0.56.0, which requires `go 1.26.0`, which is a toolchain bump across every workflow, Dockerfile and doc — not a dependency bump, and not something to hide inside a scanning PR. Neither CVE is reachable from this code (govulncheck green; `pkg/ssh` uses x/crypto/ssh for key marshalling only, not for a connection or its mux), so it is scheduled work rather than an incident.

Trivy will open two alerts for them on the first run after this merges. That is the intended behaviour and #261 is where they get closed — the point of #258 was never "no alerts", it was "no alert that nobody can act on".

scttfrdmn:
Corrected the CodeQL half. The first attempt used `strconv.Quote`, which escapes newlines perfectly well but is not a barrier `go/log-injection` knows about — so all four alerts stayed open. I found that by querying the alerts on `refs/pull/260/merge` rather than reading the green CodeQL job, which only says the analysis ran.

Per `LogInjectionCustomizations.qll` the modelled barriers are exactly two: a `strings.ReplaceAll` whose replaced string is `\r` or `\n`, and an argument formatted with `%q` at the format call itself. It is now `%q` at both sinks, which is also less code than the helper it replaces and puts the escaping where a reviewer looks for it.

